PR #41, opened 2016-02-21
CONTRIBUTING docs

sevki:
these are not in the wiki because these, when in the repo are special
files that change behavior in Github.

A please "check out our contributing guide" warning appears on top of pull-request
and issues forms @ Github, like so

img

Change-Id: Idb35cdb54bc9ff0cbb2564e122c138d8fea4f885


elbing:
Ok for me.

PR #44, opened 2016-02-21
Remove envsetup and change instructions in bootstrap.sh

sevki:
It's a huge problem causer and I think it's why people keep not catching problems
with the build. We're going to move to a world with the new build in which environment
variables get set correctly. For now, let's stop causing trouble for ourselves.

I had a day of false positives with harvey builds because of this damned script. envsetup makes it way too easy
to move to a new tree, forget you have the wrong variables set, and be building in your old tree. It's horrible.

Change-Id: I644d6772f0923006798a51f007283f2eda0b286a

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -18,10 +18,11 @@ mkdir -p tmp
 
 cat <<EOF
 
-Now just set your environment variables and build:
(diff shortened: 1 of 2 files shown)


sevki:
already in

PR #48, opened 2016-02-21
Delete Buglist and AUTH_HOWTO

sevki:
are these still necessary?

AUTH_HOWTO moved to the [wiki](https://github.com/Harvey-OS/harvey/wiki/Auth-How-to)


elbing:
Ok for me. It comes from ancient times...

PR #54, opened 2016-02-22
Fixing again ninep building issues ("lionkov").

elbing:
I think this should get master branch by default, --remote flag
and "new commits" detected from Ron's repo.


rminnich:
Looks find to me assuming it works.

elbing:
Only way to test is cloning repo, applying patch and see if after bootstrap.sh there's a util/ufs.

sevki:
no util/ufs

elbing:
Fuck... any idea? if you go to ninep path and do git branch -a you will see it's not in master branch, an then Ron's changes to fix Go path problem don't work.

PR #59, opened 2016-02-26
riscv. Well, this is pretty messed up.

rminnich:
Why is all this stuff in there in addition to my commits?


rminnich:
ah it actually seems ok. BTW, riscv won't build, of course; I'm going to bring in bits from coreboot to make it work.

elbing:
I'm impatient to put my hands on this. But other tasks are claiming for me.

PR #60, opened 2016-02-27
Changes for RISCV: de-amd64-size config and tools, start fleshing out riscv tree.

rminnich:


elbing:
Nice!, specially syscalls build.json. You should pull from master, seems that any changes passed while you were working on this.

PR #62, opened 2016-02-27
riscv: relocate some files that are architecture-dependent; fix mksys.go for riscv syscalls

rminnich:


elbing:
This doesn't build, Ron.

PR #64, opened 2016-02-28
More work on riscv

rminnich:


elbing:
This is more the final cleaning of json mess from "all in ramfs" days a year ago. Nice, Ron. This should have been done months ago. Thanks.

PR #68, opened 2016-02-29
Further riscv changes. User space now builds completely.

rminnich:


elbing:
Why did you change sys/src/cmd/aux/realemu/loadcom.c/json files? I think "-I ./ " is more elegant than specifying entire path for a header. In fact realemu won't be used under riscv if you force a amd64/include header.

rminnich:
Because it can be used under riscv. It's just an emulator. It should run on any CPU. And if there are ever riscv CPUs with pcie slots than we will need to be able to run the options ROMs. 

The -I / makes more sense because we need to name the architecture explicitly and you can't do that with just a ..

Using -I / for this one case also makes it easy to move the file around in the tree and not have to change the header all the time.

elbing:
sold out!.

rminnich:
I still don't see the problem.

elbing:
no problem, you sold it to me :-)

PR #72, opened 2016-03-01
More rearrangement to get riscv to build; add a few files.

rminnich:


elbing:
How will I reboot remotely my real server since now?
Why devpci is not available from port.json? RISCV hasn't pci?

PR #73, opened 2016-03-01
corrected riscv support

rminnich:


elbing:
Could you move rebootcmd to amd64 if it's incompatible?

rminnich:
rebootcmd is portable. I just am not ready to implement it in riscv yet. It should not move.

elbing:
Just not going upstairs three floors to reboot my server, please move it temporary to amd64, like uahci and others.

rminnich:
oh wait I screwed up. I took out too much code. I will put back the if (argc == 0) exit(0) and it will work on all systems. 

What I took out was the reboot with a file name, which does not work.

PR #74, opened 2016-03-01
Move things around, but don't break rebooting with no file name.

rminnich:
reboot still works.


rminnich:
this builds and boots and the reboot command still works.

elbing:
LGTM

PR #75, opened 2016-03-01
Fixed crash in stats

keedon:
stats was trying to update a menu in protected memory.
It's now copied to writeable memory at start

Signed-off-by: Keith Poole <keith.poole@gmail.com>


rminnich:
LGTM, this is a reasonable way to get writeable strings that is not gcc-specific.

PR #78, opened 2016-03-03
riscv: it can now link. It has a main.

rminnich:
Note we don't need assembly since firmware starts a kernel with
paging set up already.


elbing:
Amazing. Well you're the master. I'll read some docs about riscv.
LGTM

PR #79, opened 2016-03-03
Add note about frame pointer.

rminnich:


sevki:
does this even need a +1? 
LGTM

PR #81, opened 2016-03-04
more riscv bits. We may have printing to a console soon.

rminnich:


elbing:
LGTM. And me without the f** native toolchain for amd64, argh :-(

elbing:
out of date. We removed unused variables thanks to Gerard...

PR #83, opened 2016-03-04
Removed [-Woverflow] warnings

grd:
I do question btw why var `h->bodylen` and `stop` which are type
uint32_t overflows with ~0UL.

io.c:30:16: warning: large integer implicitly truncated to unsigned type
[-Woverflow]
   h->bodylen = ~0UL;

parse.c:456:10: warning: large integer implicitly truncated to unsigned
type [-Woverflow]
   stop = ~0UL;


elbing:
Try a cast (uint8_t)~0UL. About reasons, @rminnich could point you better than me, because this is about compiler internals, I think. Ask in list, Ron and John could explain it nicely.

elbing:
Oh, and you need update branch here :-).

grd:
@elbing a ~ is bitwise not. So I think that ~0UL is 0xFFFFFFFF. But uint32_t is unsigned and I am a bit surprised that it overflows. A casting with (uint32_t)~0UL makes more sense. On the other hand I am not a C expert so I could be entirely wrong ;-)

elbing:
well, uint32_t		bodylen;	/* remaining length of message body */. So uint32_t. This is because I think ~0UL is really an uint64_t integer, 0xffffffffffffffff.

elbing:
But don't suffer too much with libhttpd. There will be better tools for html parsing and modern webservers in Go or APEX. For now just to build entire system with -Werror.

grd:
@elbing if 0UL is 64 bit then the patch of 0L should also overflow. AFAIK (U)L is 32 bit and (U)LL is 64 bit.

elbing:
Ok 0L it's ok for me.

0intro:
> io.c:30:16: warning: large integer implicitly truncated to unsigned type [-Woverflow]
> h->bodylen = ~0UL;

You should just drop the L. The reason is that on Plan 9, long is always 32-bit (LLP64),
while on Unix it is the same length as a pointer (32-bit on 32-bit machines and 64-bit
on 64-bit machines) (LP64).

When moving httpd.h from Plan 9 to Harvey, ulong was naturally translatated to uint32_t.
But ~0UL should also have been translated to ~0U.

elbing:
That has more sense. Thanks David.

grd:
Oops. That last one wasn't mine.

rminnich:
gdr: you can always try a git pull --rebase, but when I get really mad, I do this:
git push rminnich :riscv # blow away the branch
git push rminnich riscv # push the new one

It's kinda like -f, just way more violent.

grd:
Ron it works now. I just removed the local repo and cloned the github repo again. I think this pull request is done and can be merged now.

rminnich:
LGTM

PR #86, opened 2016-03-04
More warnings

grd:


grd:
Yes, default:; looks weird but it didn't compile with default: alone.

rminnich:
This needs to be revised. Just casting things that are const to non-const is asking for trouble. I realize fixing the declarations is painful but it's what we have to do.

rminnich:
The convention then is 
default:
\tbreak;

rminnich:
Yeah you're right I got mixed up with go again. But in any event the default:; is something I've not seen done. Just put a defaut:
break;

in there.

grd:
Okay will do tomorrow.

elbing:
Yes, you can remove acidleak.c and nm.c too. Old debugger has nothing to do now and nm is absolutely useless since we have binutils 2.26. Old nm can't read gcc objects.

grd:
The mentioned files to be deleted will be another pull request.

grd:
Next will be deleting unused functions [-Wunused-function]

rminnich:
so what are we doing with this one again? If it's not going in please close it. If it is going on then rebase it.

grd:
I don't think there is anything wrong with it.

rminnich:
You'll need to rebase then.

rminnich:
LGTM

Issue #14, opened 2015-08-13
Build with -Werror


rminnich:
We should always be able to build with -Werror

elbing:
Prepare to make casts as a wild tiger. We have tons of warnings.

sevki:
@rminnich my experience with this, as clang is stricter with what is an error breaks the builds. I don't think we can claim this right now. But I think we should close this issue and open issues for targets that aren't being built with -Werror instead of this umbrella it's not working issue.

grd:
The days are getting shorter and the evenings longer... I would like to pick up the warnings again. About Rune, the way I see it there are three possibilities. One is to make Rune int32_t, two is to add (Rune *) in front of each "L" string and the third is something else. Which is preferred?

rminnich:
we changed it to uint on amd64we changed it to uint on amd64

grd:
Yes but isn't that causing all these warnings?

rminnich:
I don't see the point. If we need the change, we need the change, and we deal with the warnings that result. That's what we did on NxM.

necktwi:
on macOS 10.12.2
```
port/getcallerpc.c: In function 'getcallerpc':
port/getcallerpc.c:16:20: error: calling '__builtin_return_address' with a nonzero argument is unsafe [-Werror=frame-address]
  return (uintptr_t)__builtin_return_address(2);
                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
port/getcallstack.c: In function 'getcallstack':
port/getcallstack.c:20:21: error: calling '__builtin_return_address' with a nonzero argument is unsafe [-Werror=frame-address]
   pcs[0] = (uintptr)__builtin_return_address(2);
                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~
port/getcallstack.c:22:21: error: calling '__builtin_return_address' with a nonzero argument is unsafe [-Werror=frame-address]
   pcs[1] = (uintptr)__builtin_return_address(3);
                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~
port/getcallstack.c:24:21: error: calling '__builtin_return_address' with a nonzero argument is unsafe [-Werror=frame-address]
   pcs[2] = (uintptr)__builtin_return_address(4);
                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #87, opened 2016-03-05
Removed [-Wunused-function] warnings #14

grd:


rminnich:
Overall: let's slow down here a bit. You want to be really careful. The noauth stuff in cpu should definitely stay, it's useful when you do a new port. But we don't do #if whatever type compilation in plan 9, and you really don't what that in there by default. You don't even want to make it easy to compile in: it's a security hole. But we've used it to effect on new ports. So how do we keep it and avoid the warnings? Just wrap the functions in #if 0 with appropriate warning? I'm fine with that. 

But the wholesale deletions bother me a bit.

Think hard before you approve this, everyone. Really, really hard.

grd:
Oops. I only removed the dead code. I thought it would give some remarks but this is quite a lot. I agree, let's slow it down a bit.

elbing:
Libsec for now it's needed, at least for drawterm and SSL v2 in kernel. Unless we improve dratwerm. For TLS in kernel too. Vendorizing libz was a first step for better crypto in Harvey, but now Rafa is working on ZFS, so this will have to wait.

elbing:
Seems like hexdumps was a common exercise for debugging or univerity students...

elbing:
Well, most of this is ok, just pay attention to Ron's comments for things tha can't be so in stats, prep and cpu. And all of that programs that will be replaced for better versions in go or posix in future (imapd, httpd, cifs...). Nice carnage :-).

rminnich:
LGTM
it's better than it was but I don't like the cast of const char *s to char *. 

oh well. It's better.

rminnich:
now it has conflicts.

grd:
Sorry, I have to do this again.

PR #93, opened 2016-03-07
The strtoul pointer casts are set wrong.

grd:
The strtoul pointer casts are set wrong. There are more strtoXY casts
set wrong. These will all get a separate CR. Issue #14.


grd:
Never mind, there are four of them more. I will group them all together in this CR.

grd:
Merged.

elbing:
I fixed str* declarations in APEX headers to be fully compliance with POSIX 2008. But we should remind that in fact we are using the funcs (like this) in libc, no more wrappers for this in libap.a, so ok for me.

LGTM.

PR #99, opened 2016-03-19
Removed ppp

grd:


elbing:
@rminnich ppp gone? For me ok, almost none is using modems these days and this should be upgraded to better thing agree with time line. Perhaps in Go or something like that in future if it's needed.

rminnich:
LGTM

PR #100, opened 2016-03-20
Added sbrkmerge debug message

keedon:
This is where the error is for mahjongg

### DIFF
--- a/sys/src/9/amd64/qmalloc.c
+++ b/sys/src/9/amd64/qmalloc.c
@@ -115,7 +115,7 @@ static	Lock		mainlock;
  * From libc malloc.c to *draw devices
  */
 
-static int		sbrkmerge(void*, void*);
+static int  sbrkmerge(void*, void*);
 static void poolprint(Pool*, char*, ...);
 static void ppanic(Pool*, char*, ...);
 static void plock(Pool*);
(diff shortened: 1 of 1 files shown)


elbing:
This is for avoid kernel crashing, no? or does it fix entire error?

sys/src/9/amd64/qmalloc.c
```diff
@@ -136,7 +136,7 @@ static Pool pimagmem = {
 	.quantum=	32,
 	.alloc=		malloc,
 	.merge=		sbrkmerge,
-	.flags=		0,
+	.flags=		POOL_PARANOIA | POOL_VERBOSITY | POOL_DEBUGGING | POOL_LOGGING,
```
rminnich: This should not be the default, so you're going to have to remove it from this commit.

sys/src/9/amd64/qmalloc.c
```diff
@@ -154,8 +154,11 @@ sbrkmerge(void *x, void *y)
 	uint32_t *lx, *ly;
 
 	lx = x;
-	if(lx[-1] != 0xDeadBeef)
+	if(lx[-1] != 0xDeadBeef){
```
rminnich: This looks fine however.

rminnich:
:-1: 
You need to remove the setting of default attributes. It can't go in in this form. The message is ok however.

PR #103, opened 2016-03-22
cmux: work in progress.

rminnich:
This is rio with all windowing bits removed.

This is the long hoped-for console mux. We'll see how it goes. I'm just giving it a shot.

For now I've left limited mouse support, because you can have a mouse ina  console on harvey.
It's /dev/mouse and it's text.

This will be amusing.


elbing:
Better to merge this if all is ok to work together on it.

elbing:
LGTM

PR #105, opened 2016-03-24
libthread: fix varargs problem.

rminnich:
Another leftover from the Giant Conversion to standard calling sequences.

### DIFF
--- a/sys/src/libthread/exec.c
+++ b/sys/src/libthread/exec.c
@@ -84,6 +84,27 @@ procexec(Channel *pidc, char *prog, char *args[])
 void
 procexecl(Channel *pidc, char *f, ...)
 {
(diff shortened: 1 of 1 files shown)


sys/src/libthread/exec.c
```diff
@@ -84,6 +84,27 @@ procexec(Channel *pidc, char *prog, char *args[])
...
+	 * allocates extra space anyway -- often realloc just returns with nothing changed.
+	 * And, finally, usually the number of args is very small.
+	 */
+	va_list a;
+	char **args = nil;
+	char *arg;
+	int argc = 0;
+
+	va_start(a, f);
+	for(;;){
```
dancrossnyc: Use a do {} while(argc != nil); here?

Issue #88, opened 2016-03-06
Running mahjongg game panics kernel


myk321:
#### Expected behavior
Running mahjongg in a rio window should result in the mahjongg game being displayed.

#### Actual behavior
rio freezes and the following error message is displayed, both over the rio screen and in the QEMU window:
read fault fail, no segment, pid 82 addr 0x0 pc 0xfffffffff0150365
qlock with ilockdepth 1, stack: k:1df7a3 k:1da556 k:155220 k:1545de k:154382 k:138f90 k:20a39d k:20a1fb k:20afeb k:20b999 k:1f5755 k:14114b k:140ac5 k:1e76b2 k:1e78f3 k:152b4d k:139695
panic: cpu0: cpu0: ilockdepth 1, last lock 0x0 at 0x0, sched called from 0xfffffffff014a1b5

#### Steps to reproduce
- Start harvey and rio
- Open a window
- At the command prompt enter: mahjongg
- Above error message is displayed

#### QEMU output
As above

rminnich:
Would you be willing to do a little kernel work? You can run the kernel under qemu and get a backtrace for us. We could use the help.

rminnich:
oh, and, thanks for trying this and reporting the problem!

myk321:
I inserted some debug statements into the code and got the following:

In mahjongg.c, around line 73 we see:
	fd = open(path, OREAD);
	if(fd < 0) {
		fprint(2, "cannot open image file %s: %r\n", path);
		exits("image");
	}
	img = readimage(display, fd, 0);
the program never returns from this readimage.

If we insert a fprintf we get the message:
	screen id in use
and see we were about to read:
	/sys/games/lib/mahjongg/backgrounds/default.bit
This file exists and is in a compressed format.

readimage() is in /sys/src/libdraw/readimage.c:15
On /sys/src/libdraw/readimage.c:31
we see that:
	if(memcmp(hdr, "compressed\n", 11) == 0) {
		return creadimage(d, fd, dolock);
Since the image is compressed we progress to:
	/sys/src/libdraw/creadimage.c:15

In creadimage() we see
	if(d){
		if(dolock) {
			lockdisplay(d);
		}
		i = allocimage(d, r, chan, 0, 0);
		setmalloctag(i, getcallerpc());
		if(dolock) {
			unlockdisplay(d);
		}
		if(i == nil)
			return nil;

Execution does not return from 
		i = allocimage(d, r, chan, 0, 0);
[...]

PR #109, opened 2016-03-27
Fix for panic in Mahjongg #88

keedon:
It looks like a bad interaction between the malloc pool and the image pool
when allocating images.  I've removed the image pool& it works.

Can someone check this out - it works, but I'm a bit twitchy I've missed
something. c.f. memimagemove

### DIFF
--- a/sys/include/pool.h
+++ b/sys/include/pool.h
@@ -51,7 +51,6 @@ extern int	poolcompact(Pool*);
 extern void	poolblockcheck(Pool*, void*);
 
 extern Pool*	mainmem;
-extern Pool*	imagmem;
 
 enum {	/* flags */
 	POOL_ANTAGONISM	= 1<<0,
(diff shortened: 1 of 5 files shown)


elbing:
I'm not sure, specially of replacing poolalloc for malloc. Poolalloc must be fixed if it's wrong, but it's a good work. Qmalloc was fixed months ago, we should have it always aligned with libc/port/malloc.c if we change anything in it. Anyway, pool mess in qmalloc kernel was included when I brought back screen and vga in September. Originally wasn't there. Surely my mistake, but for an arcane reason I can't remember, I found some undefined things that required including pool mess in it (I remember a talk with Ron about this) that originally was in old alloc.c. I'm not sure if it was due to some sources now removed or changed. So removing it now couldn't be wrong at all, now it seems as it was before September 2nd.

Look at: [Original qmalloc.c](https://github.com/Harvey-OS/harvey/commit/da1fc778d8ec321fb85a6c044bfbc0ac0bc81ef7)

rminnich:
SGTM

rminnich:
If this works for you go for it.

elbing:
LGTM

PR #113, opened 2016-04-09
Prototype fdmux.

rminnich:
This is an fd mux, intended for harvey and akaros.

A program can attach #< somewhere in its namespace. It can then open data and data1.
It can spawn a program with data1 as 0, 1, and 2.

From that point, it can read/write data and the child will see this as 0, 1, and 2.

At some point, the parent might spawn another child, in a different pgrp, with these as 0, 1, and 2.
fdmux will only let the child with the matching pgrp do IO. The other one will block.

This is an attempt to allow ptty-like operation with a more modern design. It is a start. Testing comes next.


elbing:
Good! LGTM

sevki:
LGTM

PR #116, opened 2016-04-11
Get gdb server building again

rminnich:
Make it a standard buld component too. Let's see if we can help alvaro!


elbing:
LGTM

PR #117, opened 2016-04-12
Rework fdmux

rminnich:
It now supports m (master), cons, and consctl and is intended to be
bound MBEFORE to /dev. This way, child processes that try to mess with
cons/consctl won't break things.

Now for ^Z ...


elbing:
LGTM

PR #121, opened 2016-04-17
Make dup match Linux dup2.

rminnich:
as part of this fix, remove sys.h from the tree.
It got added March 10 for reasons I don't
understand (the commit message is not at all helpful in this regard).

sys.h should not be added to the tree. It's a GENERATED file.
We should never add generated files to the tree.


elbing:
LGTM

rminnich:
Already done.

PR #122, opened 2016-04-18
Take the new build out.

rminnich:
It's not working correctly and the hack to fix it is breaking other things.


rminnich:
LGTM

rminnich:
BTW, by this I mean: putting a generated file (9sys.h) into the git repo is the WRONG way to fix whatever the new build tool is doing incorrectly (not generating that file? I have no idea). Sorry :-(

PR #124, opened 2016-04-18
put sys.h back in to .gitignore. As part of the mistaken addition of sys.h, it was removed from .gitignore. Put it back.

rminnich:


sevki:
shouldn't this also delete `sys.h` from the tree?

rminnich:
sys.h is already deleted from the tree, earlier.

elbing:
LGTM

PR #128, opened 2016-04-23
More deanon.

rminnich:
I don't know the right way to do the mouse bits, see what you think.

But it works, or seems to.


elbing:
This commit is included in 129. So it works in real hw.
LGTM.

PR #129, opened 2016-04-23
Enable clang builds.

rminnich:
Finally, at long last, we can try clang on the kernel,
once we remove these gcc options.

And, wow, it's fast.


sys/src/9/port/devcons.c
```diff
@@ -757,7 +757,7 @@ conswrite(Chan *c, void *va, int32_t n, int64_t off)
 			rebootcmd(cb->nf-1, cb->f+1);
 			break;
 		case CMpanic:
-			*(uint32_t*)0=0;
+			*(volatile uint32_t*)0=0;
```
elbing: wow!

sys/src/9/port/initcode.c
```diff
@@ -43,6 +43,8 @@ startboot(char *argv0, char **argv)
 	bind(ec, env, MAFTER);
 	bind(e, env, MCREATE|MAFTER);
 	bind(s, srv, MREPL|MCREATE);
+	write(1, "HI\n", 3);
+	write(1, boot, 10);
```
elbing: Why this? init9 related?

elbing:
Don't suffer about mouse, for me it's ok (you found buttons missing mess in struct). I wonder about init9.
"vartracking" gcc flags are not needed. I can't remember why are they there, I think when Giacomo put dwarf option too.
LGTM.

rminnich:
Alvaro, thanks for checking this. That "HI" print was debug code I left in by mistake! NIce catch!

PR #130, opened 2016-04-24
Remove the probe for ISA cards.

rminnich:
The index -1 was supposed to key a search or ISA cards. ISA is dead. I removed the
code but missed the probe of -1. Nothing broke.

This panic was not found until we started building with clang and the kernel
panic'ed. We ran for a year with this bug.


elbing:
Wow hard to find. Good catch!
LGTM

PR #131, opened 2016-04-24
clang: start to resize libmach

rminnich:
Remove much stuff we don't need. More will be removed later.


elbing:
LGTM

PR #132, opened 2016-04-25
clang: add sys/src/9/amd64/clang.json

rminnich:
Until we figure out how to special-case flags for compilers (or we get that in the new build tool)
just use this file for building the kernel. We don't want xmm usage in the kernel for now.

To use
CC=clang build sys/src/9/amd64/clang.json

### DIFF
--- a/sys/src/9/amd64/clang.json
+++ b/sys/src/9/amd64/clang.json
@@ -1,0 +1,133 @@
+{
+	"cpu": {
+		"CFlags": [
+			"-mno-implicit-float"
+		],
+		"Env": [
+			"CONF=cpu"
+		],
+		"Include": [
+			"core.json",
+			"../386/386.json",
+			"../ip/ip.json",
+			"../port/port.json"
+		],
+		"Kernel": {
+			"Config": {
+				"Code": [
+					"int cpuserver = 1;",
+					"uint32_t kerndate = 1;"
+				],
+				"Dev": [
+					"acpi",
+					"arch",
+					"cap",
+					"cons",
+					"coreboot",
+					"draw",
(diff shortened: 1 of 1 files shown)


elbing:
Ok, perhaps we should add -mno-implicit-fp for gcc, if we want to keep disabling xmm registers.

elbing:
LGTM

dancrossnyc:
Another option is to make a little shell script to wrap the compiler and
stuff compiler-specific options in here.

elbing:
I thought in it, but given the facility provided in build.go, an alternative FLAGS file in sys/src or just a custom kernel json like Ron is suggesting is a fine provisional way. Once native toolchain could build the system, and llvm would be ported, probably we will never care about it anymore. But there's still a lot of work to do. Another script is another shell dependency more.

rminnich:
A script starts to feel like a return to the bad old days of configure and dltool. It is easy to extend our current tool to support per-compiler options, but I want to see if we're going to go with the new build tool or abandon it before I do any more.

dancrossnyc:
On Mon, Apr 25, 2016 at 10:17 AM, ron minnich <notifications@github.com>
wrote:

> A script starts to feel like a return to the bad old days of configure and
> dltool. It is easy to extend our current tool to support per-compiler
> options, but I want to see if we're going to go with the new build tool or
> abandon it before I do any more.
>
I was actually thinking of what Russ Cox did with plan9port and his '9c'
script: it papers over the differences between various compilers and gives
one the illusion of cross-platform consistency.

But another file is of course fine as well. And it's exciting that things
are working with clang now!

        - Dan C.

PR #136, opened 2016-04-26
Add the ability to extend options for a tool.

rminnich:
It's a bit of a kludge, but it will hold us until we get the new build tool.

There is a new .json file entry, "ToolOpts". To use it, name the full path
of the tool, and if you are using that tool, the opts get added.

So, for example, for /usr/bin/clang, you can see we do this:
"ToolOpts": {
	"/usr/bin/clang": [
		"-mno-implicit-float"
	]
}

Tested and correctly invoking clang. The built kernel works, modulo the ongoing
bug with failing to get dhcp back (frequently).


elbing:
LGTM

PR #142, opened 2016-04-30
deanon libcontrol

rminnich:
Not really building without plan9-extensions but this is a step.


elbing:
LGTM

PR #145, opened 2016-05-02
Fix an error resulting from the deanon work.

rminnich:
The expression could never be nil. The original was testing against
the value of p in the lock, i.e. whether another process wanted to use
the pool and we got there first.

This is pretty amazing stuff, in that it works at all, but the use of anon
struct members made it even more confusing. Overall, I think they're a bad deal.


elbing:
LGTM

PR #149, opened 2016-05-06
Bring back acid

rminnich:
we need a debugger, and acid might be a reasonable standin until
somebody does gdb. I can't stand to look at gdbserver any more.

This is just an initial import.


elbing:
I'll trust in what you're doing, because gdb won't come soon. Even with a native gcc, I (and first David Hogan 13 years ago) never got gdb properly working in Plan 9.

+1

PR #154, opened 2016-05-13
build: testing using macros and loops

sevki:
this needs more work, mainly the ports/*.c files and their relative includes, and the fact that the branch of build that can build this hasn't been merged in to master

[instructions](https://gist.github.com/sevki/8a3d4b80cafb0c6009cc6fab7528ab47) to build


BUILD
```diff
@@ -3,9 +3,10 @@ objcopy(
 	deps=[
 		"//sys/src/9/amd64:harvey"
 	],
-	in="elf64-x86-64",
-	out="elf32-i386",
+	infile="elf64-x86-64",
```
sevki: this has changed because in is a keyword in build syntax

elbing:
what's scan-build?

sevki:
@elbing [clang's static analyzer ](http://clang-analyzer.llvm.org/scan-build.html)

rminnich:
ok will review tonight, give me some time.

sevki:
ok. thanks for the review :smile:

sevki:
build breaks because of Harvey-OS/ninep#8

bootstrap.sh
```diff
@@ -13,6 +18,6 @@ cat <<EOF
 # We support RISC-V, but the default is x86_64 (which we call amd64 for historical reasons):
 export ARCH=amd64
 # And build:
-./util/build
+./util/build -v //:kernel
```
rminnich: nope. you can't do this yet  because blaze can't replace build. 

for now, blaze should be additive, it should not remove anything already there. This command won't produce a working tree, and the other command line will.

bootstrap.sh
```diff
@@ -5,11 +5,12 @@ git submodule update
 
 GOBIN="$(pwd)/util" GOPATH="$(pwd)/util/third_party:$(pwd)/util" go get -d harvey/cmd/... # should really vendor these bits
 GOBIN="$(pwd)/util" GOPATH="$(pwd)/util/third_party:$(pwd)/util" go install github.com/rminnich/ninep/srv/examples/ufs harvey/cmd/...
- 
-echo Downloading the build tool...
-curl -L http://sevki.co/get-build -o util/build
-chmod +x util/build
- 
+
+if [ $OLD_BUILD != true ]; then
```
rminnich: for now, we need both tools. Also you're still extracting into build, not blaze. So we won't get a complete tree.

This was why I had to remove the new build tool last time: it broke stuff.

Let's take this a bit more slowly. Can you submit JUST a PR that adds all the BUILD files? This is too much to digest. For me, anyway.

sevki:
@rminnich why is @github showing you an outdated diff I don't know but this is the correct one https://github.com/Harvey-OS/harvey/pull/154/files#diff-3af436e6760a0c5f1e3d16a3384853a8

PR #155, opened 2016-05-13
Fix the clean rule

rminnich:
It now blows away $ARCH/bin without thinking hard (because, stupidly, amd64/bin includes
some rc scripts, i.e. not binaries -- wtf?) and all the .a's in lib. Then, to undo
any possible damage, it does a checkout of $ARCH/bin


elbing:
+1

PR #157, opened 2016-05-13
build: change environment variable for CC to HARVEYCC

rminnich:
travis seems to want to override it.


sevki:
it does it because of this [line](https://github.com/Harvey-OS/harvey/blob/master/.travis.yml#L5) if you just added clang there it owuld work fine

PR #159, opened 2016-05-16
clang: more cleanup

rminnich:
Lots left to do, but you can only stand to do this a little at a time :)


elbing:
LGTM. It builds.

PR #160, opened 2016-05-21
Added cron back into startup script

keedon:
I've also stopped the warning message from appearing on the console.
It's still appears in the /sys/log/cron file unless you run:

sudo chown -R sys:sys cron/sys

before starting harvey.


sys/src/cmd/auth/cron.c
```diff
@@ -293,7 +293,7 @@ readalljobs(void)
 			    !(d[i].qid.type & QTDIR))
 				continue;
 			if(strcmp(d[i].name, d[i].uid) != 0){
-				syslog(1, CRONLOG, "cron for %s owned by %s",
+				syslog(0, CRONLOG, "cron for %s owned by %s",
```
rminnich: So, wait, you're now writing to stdin?

rminnich:
LGTM

sys/src/cmd/auth/cron.c
dancrossnyc: No, that first argument to syslog is a flag specifying whether to write to #c/cons or not.

PR #161, opened 2016-05-21
Removed dependency on awk in cpurc

keedon:
bootstrap.sh changed to include instructions to make drawterm
work correctly.


elbing:
LGTM

dancrossnyc:
Use 'field' instead of 'print1'? That's what I wrote it for. :-)

keedon:
Didn't know about field, I'll change it

PR #164, opened 2016-05-28
icc changes. This mostly builds. We need to figure out how icc expesses returns_twice.

rminnich:


elbing:
Fortunetely, we haven't too much inline asm. Usually the convention for gcc and clang is also __asm__.

elbing:
LGTM

PR #168, opened 2016-05-31
Changed qmalloc to use pool.c from libc

keedon:
Rationalised to simplify adding a slab allocator at a later time.

### DIFF
--- a/sys/src/9/amd64/qmalloc.c
+++ b/sys/src/9/amd64/qmalloc.c
@@ -317,7 +317,7 @@ sbrkmerge(void *x, void *y)
 
         lx = x;
         if(lx[-1] != 0xDeadBeef)
-               abort();
+               panic("invalid memory area in sbrkmerge %x %x", x, y);
 
         if((uint8_t*)lx+lx[-2] == (uint8_t*)y) {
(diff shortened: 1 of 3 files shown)


elbing:
what shall we do... (popular song) with this?

keedon:
No one seems keen, so I'll shelve it for now

elbing:
@rminnich  usually says: "if it can be out of kernel, it must be out of kernel". I think you said it doesn't improve too much the system (faster), but if numbers are similars and it could be out... I vote for using pool from libc. Qmalloc (jmk work?) could be rescued later as a subproject or something, I think we talked about it months ago and someone said it was an unfinished work.
@rminnich, your wisdom, maybe I'm wrong.

rminnich:
What's going on with this one. Just awaiting another tester?

sevki:
I tried to carry this PR but as it turns out @keedon has deleted his fork and re-forked, hence there is no way this can get merged and is effectively obsolete, that's why I'm closing this. 

If anyone else wants to carry this, download and apply the patch @ https://patch-diff.githubusercontent.com/raw/Harvey-OS/harvey/pull/168.diff

PR #171, opened 2016-06-05
Replacing awk for field

elbing:
setrc script is called from fshalt.


dancrossnyc:
Hmm. I think you might have to fetch and rebase against master; I can't seem to be able merge directly.

elbing:
done, I had outdated my local master.

PR #172, opened 2016-06-06
Up/down keys now move cursor up/down one line

keedon:


rminnich:
ah, hmm. You've just made a huge change in acme behavior. Let's talk about this.

fuchicar:
I'm agree to add this behavior (I will use it), but maybe it should be activated with a flag or with an internal command like when you activate autoindent with "indent on".

cratuki:
This reminds me of a change that redhat made to the default behaviour of vim in the early days. They set up the default profile of vim so that you could use cursor keys to navigate around when in insert mode. From this, a generation of users who missed the point of vi.

At the same time, something that makes Harvey exciting is that it is creating pragmatism in a community that's not known for it.

Harvey might become this: a good bridge by which people from unix backgrounds can migrate into plan9. Is that Harvey's mission? If so, is this change in the spirit of that?

rminnich:
oh let's try it
LGTM

elbing:
@cratuki , one of the goals, but not the only one. But yes, one of the big walls to break is that weird fame of unusable for what 9 systems are sadly known. But many other things are on fly.

keedon:
@rminnich @elbing @cratuki  I'm wondering if I should create a new "acme" (nadir?) and leave classic acme as it is.  There are quite a few changes I'd really like to make, but I don't want to alienate everyone??

elbing:
Not bad idea, I modified rio in past and called rio2 (privately) to not breaking replica. Please, do it. Then changes on it could be introduced lab style, almost without discussion.

keedon:
I'll update this tonight & retry

PR #175, opened 2016-06-07
This ones for you Dan, I hope it helps.

rminnich:


elbing:
LGTM

aarch64/bin/auth/disable
```diff
@@ -0,0 +1,5 @@
+#!/bin/rc
```
dancrossnyc: Hmm; we need this for each arch? Surely this is a candidate for /rc/bin/auth and a 'bind -a /rc/bin/auth /bin/auth' ?

dancrossnyc:
Ron beat me to it. :-)

rminnich:
So dan the clear.json script does an rm on /$ARCH/bin but then has to do a checkout to get the scripts back. It's because the scripts should not even be there.

dancrossnyc:
Ha! Sad Trombone....

dancrossnyc:
Hmm. Travis seems to have completed successfully, but Github hasn't picked up on that yet. Is there some way to force it to check?

dancrossnyc:
(I think I found a way.)

PR #176, opened 2016-06-08
Use /bin/sh instead of bash for more portability

raylai:
I did a cursory glance of the scripts, I didn't see any bashisms. Improves build on OpenBSD.


dancrossnyc:
Thanks! Merged in.

PR #177, opened 2016-06-08
Fixing order of elements for ld

elbing:
This should fix BSD issues, and it's the right way
to do an static compilation.


elbing:
Nobody cares about this? I know old build is closer to be retired, but until now...

rminnich:
LGTM

PR #181, opened 2016-06-09
amd64/sdata.c: Make buf[] in atadrive() static.

rminnich:
The atadrive function was declaring a local 512 byte buf, i.e. a
stack variable. Big arrays like this on kernel stack are almost always
a bad idea. In this case we were getting a page fault on NULL while probing.

I'm pretty sure this function is single threaded, i.e. only one invocation
is ever active at a time. So I just made buf[] static. The sdata panics
promptly ceased.

If you ever want to make this function reentrant you'll need to use malloc to
allocate buf[] and be sure to free it on all the (many) exits.

### DIFF
--- a/sys/src/9/amd64/build.json
+++ b/sys/src/9/amd64/build.json
@@ -70,6 +70,7 @@
 					"usbehci"
 				],
 				"Sd": [
+					"sdata",
 					"sdiahci"
 				],
 				"Uart": [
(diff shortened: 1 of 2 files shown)


elbing:
I'll pull rebase my PR when this was merged. I added some chipsets and moved it to 386 dir, also back devpnp for isa due to i440FX model (qemu and others).
LGTM

PR #184, opened 2016-06-10
Update bootstrap.sh

rminnich:
We don't yet tell anyone to use it because we're still importing BUILD files.

Author:    Sevki <s@sevki.org>


sevki:
LGTM

PR #185, opened 2016-06-10
sys/src/libmach/BUILD: Change the target name in all

rminnich:
It makes more sense: you say
blaze //sys/src/libmach:all

### DIFF
--- a/sys/src/lib9p/BUILD
+++ b/sys/src/lib9p/BUILD
@@ -1,7 +1,7 @@
 load('//sys/src/FLAGS', "LIB_COMPILER_FLAGS")
 
 cc_library(
-	name = "lib9p",
+	name = "all",
         copts = LIB_COMPILER_FLAGS,
         includes=[
             "//sys/include",
--- a/sys/src/libString/BUILD
+++ b/sys/src/libString/BUILD
@@ -1,7 +1,7 @@
 load('//sys/src/FLAGS', "LIB_COMPILER_FLAGS")
 
 cc_library(
-	name = "libString",
+	name = "all",
         copts = LIB_COMPILER_FLAGS,
         includes=[
             "//sys/include",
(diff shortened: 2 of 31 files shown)


elbing:
LGTM, update branch.

sevki:
this won't work for a bunch of reasons, so I'll list them all
firstly I forgot to add the redeclreation checker after I've refactored the preprocessor so it didn't tell you that you have redeclared all multiple times in one file like [libc](https://github.com/Harvey-OS/harvey/pull/185/files#diff-fe9aa18772c9ae10e7bad20bd0ad8f7c). that has been fixed. and the binary has been updated, now when there are duplicate target declarations it the process will die with a helpful message
```
build -v //sys/src/libc:libc
Target libc is declared more than once at these locations:
	 /home/sevki/code/harvey/sys/src/libc/BUILD:459: 
	 /home/sevki/code/harvey/sys/src/libc/BUILD:448: 
```
but more importantly than that it relies on the `libfoo` becomes `-lfoo` convention. If you are compiling a binary or a library that relies on another library build will go trough all the dependencies and look for lib prefix and will add the appropriate `-l` flag to the linker options. 
this behaviour is defined [here](https://github.com/sevki/build/blob/master/targets/cc/cbin.go#L74). This allows us not having to keep seperate `-lfoo` array and a dependencies array, and keep both of them in sync. When adding `-lfoo` flags, build respects the order, so the order of dependencies is important too. 

lastly there are already `libs`, `klibs`, `cmd` and `all`, so `blaze //sys/src:all` will build all targets. they are defined defined in the [sys/src/BUILD](https://github.com/Harvey-OS/harvey/blob/master/sys/src/BUILD) file. Perhaps moving that to the root, would be a good idea so invoking `build //:all` builds everything

elbing:
Ok, for blaze questions I'll inhibit for now. Sorry for the noise and lgtm.

rminnich:
Seems like a bad idea, I'll drop it now. I'm not thinking clearly anyway, blaze is so smart I can just always say "build world" and it only builds what's needed.

PR #187, opened 2016-06-10
blaze: just make the kernel depend on //sys/src/cmd:cmds

rminnich:
blaze is so much smarter than our old build we can just do it this way.
One issue: this fails with a very unclear diagnostic, I have to run it twice.
Help?

### DIFF
--- a/sys/src/9/amd64/BUILD
+++ b/sys/src/9/amd64/BUILD
@@ -1,4 +1,6 @@
 load("//sys/src/FLAGS", "LIB_COMPILER_FLAGS")
+load("//sys/src/cmd/BUILD", "CMDS")
+load("//sys/src/BUILD", "LIBS", "KLIBS")
 
 CORE_SRCS = [
     "entry.S",
@@ -54,7 +56,6 @@ PORT_SRCS = [
     "//sys/src/9/port/devdup.c",
     "//sys/src/9/port/devenv.c",
     "//sys/src/9/port/devfdmux.c",
-    "//sys/src/9/port/devkexec.c",
     "//sys/src/9/port/devkprof.c",
     "//sys/src/9/port/devmnt.c",
     "//sys/src/9/port/devmouse.c",
@@ -73,7 +74,6 @@ PORT_SRCS = [
     "//sys/src/9/port/devuart.c",
     "//sys/src/9/port/devwd.c",
     "//sys/src/9/port/devws.c",
-    "//sys/src/9/port/devzp.c",
     "//sys/src/9/port/edf.c",
     "//sys/src/9/port/elf64.c",
     "//sys/src/9/port/ethermii.c",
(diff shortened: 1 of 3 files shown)


sevki:
since the goal of this thing is not to generate anything in the tree, this will not work without this slight modification. 

As for the reason why this modification is necessary, dependencies install stuff to their dependants, `cc_binary` installs it's binary to `bin/foo` in everything that has defined foo as a dependency. In Unix systems this install is a symlink, in harvey this is a bind. So the `config` rule, looking at its deps knows that under `bin/` there will be `bin/echo` `bin/bind` and so on. 

This is great because we symlink whatever is being installed(binary, library, mksys generated file) to the exact hash of the dependency that we need.
For instance, when building libc
```
├── libc-623dc7b991af638b7474f6f934fbb3475bf6476d
│   ├── 9syscall
│   │   ├── alarm.s -> /tmp/build/9syscall-a3d245729eebc95682d7160018971db36a07a430/alarm.s
│   │   ├── await.s -> /tmp/build/9syscall-a3d245729eebc95682d7160018971db36a07a430/await.s
│   │   ├── bind.s -> /tmp/build/9syscall-a3d245729eebc95682d7160018971db36a07a430/bind.s
```
we can see that all these items are bound to a specific hash. As a result, we don't need to do a `make clean` type operation before every build because we know we are always using right thing version. 

Anyways, here is the "slight" modification, 

sys/src/cmd/BUILD:
```python
CMDS = [
        "//sys/src/cmd/acme:acme",
        "//sys/src/cmd/rc:rc",
        "//sys/src/cmd/ip/ipconfig:ipconfig",
        ":echo",
        ":mount",
        ":date",
        ":srv",
        ":ls",
        ":bind",
]
group(
    name="cmds",
    deps=CMDS,
)
```
sys/src/9/amd64/BUILD:
```python
load("//sys/src/cmd/BUILD", "CMDS")
config(
    name="amd64cpu",
    deps=CMDS,
    ...
)

```

rminnich:
do you want to change this CL with that modification or should I? I'm still a bit confused by all this so maybe it should be you.

sevki:
sure no problem.

sys/src/9/amd64/BUILD
```diff
@@ -377,12 +378,7 @@ cc_binary(
         ":systab",
         ":sys",
         ":inith",
-        "//sys/src/libmemlayer:libkmemlayer",
-        "//sys/src/libmemdraw:libkmemdraw",
-        "//sys/src/libdraw:libkdraw",
-        "//sys/src/libc:libkc",
-        "//sys/src/libip:libkip",
-        "//sys/src/libsec:libksec",
+        "//sys/src:klibs",
```
sevki: groups issue

sys/src/9/amd64/BUILD
```diff
@@ -525,7 +493,7 @@ cc_binary(
         "-mcmodel=small",
     ],
     deps=[
-        "//sys/src/libc:libc"
+        "//sys/src:libs",
```
sevki: groups issue

sevki:
you are getting this issue because it thinks that libs is a library called s, you need to have the libraries listed there individually, or you can do the same thing you did with cmds and have it as an array in sys/src/BUILD

sevki:
LGTM

PR #193, opened 2016-06-13
Added customisable keymapping to zenith

keedon:
See zenith.init for details - apart from the existing keyboard shortcuts, I've added:

^s^x to execute the current line
^s^f to display a fortune (just to test the external command keymapping)
^s^l to run ls -l in the current directory.

There's a minor bug with up/down keys going over a blank line, but I wanted to get this change in before I looked at it.


rminnich:
I'll never tell Rob about you.

LGTM

PR #194, opened 2016-06-14
9/amd64/BUILD: remove zp

rminnich:
zero copy pipes: maybe we want them, someday. Not now.


rminnich:
I think we don't need it.

PR #196, opened 2016-06-15
Adding needed flags for OpenBSD egcc

elbing:


elbing:
Probably clang needs the same por OpenBSD.

raylai:
clang is not in /usr/bin on OpenBSD, can the paths be removed and just the binary names used?

elbing:
It needs -nopie in final kernel linking line. But it doesn't boot.
I think without a newer binutils package it won't work on OpenBSD, too odd versions compiler<->binutils.

rminnich:
six weeks is long enough, let's let this one go.

PR #197, opened 2016-06-16
Panic if there is no MP table

rminnich:
The code assumes, from the earliest stages, that apicbase
has been set. Right now, it's only set in the mp table parser.

The result is that there are obscure failures if no _MP_ is found.

We need to drag this code base into the third millenium but at least
for now make an error that is useful, instead of just a panic on
referencing address 0x20.

THINK CAREFULLY BEFORE YOU LGTM THIS ONE.

But it at least gives me a comprehensible error on coreboot on the thinkpad.

### DIFF
--- a/sys/src/9/amd64/mp.c
+++ b/sys/src/9/amd64/mp.c
@@ -419,8 +419,9 @@ mpsinit(int maxcores)
 	_MP_ *mp;
 	PCMP *pcmp;
 
-	if((mp = sigsearch("_MP_")) == nil)
-		return;
+	if((mp = sigsearch("_MP_")) == nil) {
+		panic("NO _MP_ table");
+	}
 	if(DBGFLG){
 		DBG("_MP_ @ %#p, addr %#ux length %ud rev %d",
 			mp, l32get(mp->addr), mp->length, mp->revision);


elbing:
This is checked one time at booting when kernel talks with firmware or is accessed after that by every core? This will break boot process in machines with MP conf tables, but if you think that boot would be buggy and dangerous, it has sense to stop here. We always can warn to people who open us an issue for that, that they could change that line to be able to boot under their own risk.

I think spending time in a variable to be passed in -append kernel params is not good enough, since we need better ACPI support at the end.

rminnich:
How will this break the boot process in machines with an MP table since it only panics if there is NOT an MP table?

rminnich:
Further, if it returns from this  function early because there is no MP table, it does not set apicbase, and later it gets an obscure panic since the timer calls end up calling apicrget which derefs a null pointer.

elbing:
Heh, that if (0) comes from one year ago. LGTM.

rminnich:
OK this is good and tested. I think we should do it, because if there is no MP table the failure is really mysterious. Harvey can't run without an MP table yet.

elbing:
LGTM

PR #198, opened 2016-06-17
ACPI: add acpica source.

rminnich:
Plan 9 has avoided ACPI for a long time, but there is no
choice but to support it if we want to run on modern platforms
and laptops. At the very least, it's not possible to correctly
configure interrupts if there is no MP table, and that's increasingly
common. And it's nice to have a working power button.

This is the initial commit of acpi from Linux a26555. It does not
build. It's for reference only.


elbing:
If it's works LGTM. Does it come from coreboot?

elbing:
LGTM

PR #201, opened 2016-06-21
ACPI: move to consortium source

rminnich:
This is a WIP, but it's starting to work.

Reading the docs for ACPICA made it clear that the version in
Linux diverges enough to make reading docs hard. Possibly
the hardest part is the change in symbol names, which move from
CamelCase to camel_case. It's very hard to reconcile the names.

This series of commits will move us to building a library with
ACPICA source. So far, it's been possible to import the source
without changing it all.


elbing:
wohooo

elbing:
Ready?

PR #202, opened 2016-06-22
ACPI: get ACPICA to build.

rminnich:
This is from Intel. We can use the Intel source straight since we build as
    a library and restrictions are somewhat less applicable.

    This is from ACPI commit f39a732d2a0575b8a28b3d59f1836c7a3e4aa9cc


rminnich:
I have no idea how hub came up with that topic. It didn't even ask me. Oh well, this is for fixing ipconfig warnings. One of the new guys was gonna do that, but they appear to have run out of time.

elbing:
LGTM

PR #206, opened 2016-06-24
ACPI: single thread OS Locks/Semaphores; disable OsOverride*.

rminnich:
This is how Minix and in some cases netBSD implement these ops.

If there's only one thread then locks/semaphores
are a no nop. We can go crazy later with multithreading
if we think that's a good idea.

And the OS does not override anything, so
set the return pointer to nil or 0.

With this change, you can now add -a to your flags and
your kernel will boot with the ACPI init stuff enabled.

No more panics. It's a start.

Here's the pxelinux.cfg for booting with ACPI:

label harvey.32bit
	kernel mboot.c32
	append harvey.32bit fs=192.168.28.128 auth=127.0.0.1 nobootprompt=tcp nvram=/boot/nvram nvrlen=512 nvroff=0   -a

default harvey.32bit

Note the -a.


elbing:
Nice! LGTM. Oh I see it's now only a kernel lib.

elbing:
LGTM

PR #207, opened 2016-06-25
ACPI: add pci config supoprt and other functions.

rminnich:
This all boots now. The big problem remaining is the
antiquated Plan 9 APIC bits. The work is only beginning.
But we seem to have ACPICA. I can even see messages
from what looks like the AML parser.


elbing:
Too clever for me, there are things I don't understand too well, but basics yes. This could be a white paper for a con.
LGTM.

PR #209, opened 2016-06-26
ACPI: fix up BUILD files so kernel almost builds.

rminnich:
There were missing flags, the library names were still wrong, lots of little things.

Still not quite there but close.


elbing:
LGTM

PR #212, opened 2016-07-03
Ramdisk

keedon:
First cut at a ramdisk.
    
    There are several issues at the moment:
    
    0) It's read only - just some sample files for debugging at the moment,
    ramwrite (if you could call it) would throw an error.
    1) ls '#@/subdir' shows the README2 file, but if you try to cat it, it
    can't be found
    2) Mounting the device removes the 'w' flag from the directory


sys/src/9/port/devramfs.c
```diff
@@ -0,0 +1,180 @@
...
+#include	"u.h"
+#include	"../port/lib.h"
+#include	"mem.h"
+#include	"dat.h"
+#include	"fns.h"
+#include	"../port/error.h"
+
+#define DIR_EXTEND_SIZE 10
+
+int memcpy(void* dst, const void* src, int len);
```
rminnich: Why is this here? I think you should use memmove anyway.

sys/src/9/port/devramfs.c
```diff
@@ -0,0 +1,180 @@
...
+
+struct Directory {
```
rminnich: No lock of any kind needed? That's a bit surprising but ... 

anyway a QLock might be useful at some point.

sys/src/9/port/devramfs.c
```diff
@@ -0,0 +1,180 @@
...
+	int inuse;
+	int allocated;
+	Dirtab ramdir[];
+};
+
+struct RamFile {
```
rminnich: same question here on locking.

sys/src/9/port/devramfs.c
```diff
@@ -0,0 +1,180 @@
...
+	int length;
```
rminnich: should length really be signed?

sys/src/9/port/devramfs.c
```diff
@@ -0,0 +1,180 @@
...
+	char data[];
```
rminnich: char? uint8_t maybe?

sys/src/9/port/devramfs.c
```diff
@@ -0,0 +1,180 @@
...
+};
+
+static struct Directory *ramroot;
```
rminnich: There should not be just one of these. The reason is that each time there is an attach it can be a different name space. So it needs to be dynamically allocated in attach.

sys/src/9/port/devramfs.c
```diff
@@ -0,0 +1,180 @@
...
+         
+        c = newchan();
+        mkqid(&c->qid, (int64_t)ramroot, 0, QTDIR);
+        c->dev = devtabget('@', 0);
+        if(spec == nil)
+                spec = "";
+        buf = smalloc(1+UTFmax+strlen(spec)+1);
+        sprint(buf, "#@%s", spec);
+        c->path = newpath(buf);
+	c->mode = 0777;
```
rminnich: you might want to run clang-format on this to get indentation and such things right.

sys/src/9/port/devramfs.c
```diff
@@ -0,0 +1,180 @@
...
+	int length;
+	char data[];
+};
+
+static struct Directory *ramroot;
+
+static char* readme = "This is the kernel ramdisk.\n";
+
+static void
+raminit(void)
```
rminnich: In general this is a non-plan9 way to do this. It's a familiar way esp. if you're used to linux. But the more idiomatic way is to use a gen function; see devproc.c for an example. 

In general plan 9 drivers don't mirror a file system structure in the data structures in the driver; rather, they respond to walks and such using a gen function. It's a bit of a subtle distinction but important. You might want to look at devproc.c to see how it's done.  devproc.c is particularly interesting because procs are so dynamic, and the driver has to deal with that, even during directory reads. it also has the locking right.

sys/src/9/port/devramfs.c
```diff
@@ -0,0 +1,180 @@
...
+}
+
+static int32_t
+ramread(Chan *c, void *buf, int32_t n, int64_t off)
+{
+	if (c->qid.type == QTDIR){
+		struct Directory *dir = (struct Directory*)c->qid.path;
+		return devdirread(c, buf, n, dir->ramdir, dir->inuse, devgen);
+	}
+	// Read file
```
rminnich: and you'll have the good fun her of implementing directory reads at some point :)

sys/src/9/port/devramfs.c
```diff
@@ -0,0 +1,180 @@
...
+	struct RamFile *file = (void*)c->qid.path;
+	int filelen = file->length;
+	if (off > filelen){
+		return 0;
+	}
+	if (off + n > filelen){
+		n = filelen - off;
+	}
+	memcpy(buf, file->data, n);
```
rminnich: There are support functions to do all this. Take a look at, e.g., devarch or devproc.

rminnich:
its'a  reasonable start but you ought to look at devproc.c to see how gen is done.

elbing:
Is it possible to have this unused in main tree? it would be easy for be patched by all of us in case anyone would discover anything. I mean, using old ramdisk at boot having this.

rminnich:
LGTM
as long as it does not break anything, let's bring it in. It will make your life a lot easier.

rminnich:
let's see what other people think. 

No, this won't interfere with an existing ramdisk.

PR #214, opened 2016-07-06
ACPI work.

rminnich:
I'm not saying this is right, only that it appears to do no harm.


elbing:
Ok then.
LGTM

rminnich:
OK, just FYI, this code should not affect anyone unless they boot with -a. 

And, even if you boot with -a, it should not affect you anyway, because the acpi setup now always returns 0. . I've tested both cases.

PR #215, opened 2016-07-06
fixfmt: fix all usages of %u so that u is a type, not a modifier

rminnich:
The rest of the world uses %u to mean unsigned decimal. Plan 9
uses it as a modifier, which is arguably better, but what can we do?
You don't always get to pick the standard.

This change modifies libc/fmt so that %u is unsigned decimal, and
modifies most uses of it so the kernel almost boots. Almost. Obviously
we didn't get them all yet.

It hangs in ipconfig.


elbing:
after a year, and seeing what happend with ipconfig last months, we should import that ip stack from BSD about we were talking last summer.

rminnich:
ipconfig is fixed. It all boots now.

elbing:
LGTM

PR #217, opened 2016-07-09
Removed sdata to fix cd hang problem

keedon:
No idea why this fixes it


elbing:
I'm sure this driver breaks smp. Before Ron's fix it was hanging the machine with up = nil. After that if it's used hangs the machine in init rc.  If not used breaks walk. Time to get it gone. Bye legacy sata.
LGTM

rminnich:
you might as well remove sdata.c, and other such files, right? 
Else someone will be tempted to use it again.

PR #221, opened 2016-07-17
fixfmt: finnishing src/cmd cleaning

elbing:


elbing:
Tested ps, ls, wc, venti and fossil and others involved. Which remains me we need an utility to make all venti/fossil related installation (argh!), until ZFS could be working.

rminnich:
LGTM

NIce job. Huge effort.

PR #222, opened 2016-07-18
fixfmt: Last cleaning of %u modifier in libs

elbing:
Restoring -Werror for kernel build

(secondary: replacing memcpy in port/devramfs for memmmove
 to avoid warning in kernel build)


rminnich:
LGTM

PR #224, opened 2016-07-20
ACPI: continue the code massacre.

rminnich:
Remove $PIR support, that table's been bogus for years on most systems.
Many vendors use the same (wrong) one for many mainboards.

More ACPI fun!


elbing:
wow, you're naking pci. Perhaps now it could be more compatible and my old nforce could work. Anyway, trusting you.
LGTM

PR #226, opened 2016-07-23
ACPI: further changes, both to pci, and ACPI

rminnich:
The pci intp pin is far more useful than the intl, which seems
to be tied tightly to 8259 mode.


dmgolubovsky:
devpci: LGTM
devacpi: not familiar with this, but have no objections.
Thanks.

rminnich:
IF someone can do the user mode equivalent fixes that would be cool.

I am slowly figuring it out.

rminnich:
but still need a real LGTM :-)

elbing:
LGTM

PR #227, opened 2016-07-27
Extend 'mksys.go' to understand aarch64.

dancrossnyc:
Note: We use the Linux convention for system call number in x8.


rminnich:
LGTM

there's no LRFGTM, too bad.

PR #228, opened 2016-07-27
Fix reallocarray and use 'rm -f' when cleaning.

dancrossnyc:
In sys/src/lib.json: use 'rm -f' when removing object files.
In sys/src/libc/port/reallocarray.c: Actually make this build
under Harvey. :-)


rminnich:
LGTM

PR #233, opened 2016-07-29
Add gcc 5.2 to the build matrix.

rminnich:


elbing:
When I saw this I thought "I'm using 5.3 in debian since months and no troubles", but well... travis.
LGTM

sevki:
there is a better way of doing this. 
when you add `packages` to travis it installs the required packages for the entire matrix, instead we should use `before-script` command to run `apt-get` to install the relevant compiler. 

this will also be very important for clang static analysis as the stock travis clang doesn't have scan-build

sevki:
LGTM

PR #234, opened 2016-07-29
Fix GO9PRUN

rminnich:
Use the HARVEY variable so that we don't have to be any
particular place to run it correctly.

Add a shebang (#!/bin/bash) and set -e so some errors
get caught.

Add a port forward for 5356 so we don't get ridiculous delays on
start. CPU servers don't run auth servers anyway, but this cpu
server would hang on the auth server.

### DIFF
--- a/util/GO9PRUN
+++ b/util/GO9PRUN
@@ -1,3 +1,5 @@
+#!/bin/bash
+set -e
 
 trap : 2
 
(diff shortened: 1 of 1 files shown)


elbing:
Woohoo cpu server in kvm fast.
LGTM

[Approved with PullApprove](https://pullapprove.com/Harvey-OS/harvey/pull-request/234/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=Harvey-OS/harvey)

PR #237, opened 2016-07-30
libz: fix broken shift of negative number

rminnich:
newer clang compilers noticed this, nothing else did.

### DIFF
--- a/sys/src/libz/patch/port-inflate.sed
+++ b/sys/src/libz/patch/port-inflate.sed
@@ -1,2 +1,1 @@
-# avoid signed left shift
-s/-1L << 16/-1UL << 16/g
+/1L.*<<.*16/s//1UL << 16/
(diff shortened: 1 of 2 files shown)


sys/src/libz/patch/port-inflate.sed
```diff
@@ -1,2 +1 @@
-# avoid signed left shift
-s/-1L << 16/-1UL << 16/g
+/1L.*<<.*16/s//1UL << 16/
```
dancrossnyc: I'm not sure this is right. Shouldn't this convert the input into something like, '~0UL << 16' ?  E.g.,

s/-1L *<< *16/~0UL << 16/

?

rminnich:
Dan, I believe you are right, I will change this again.Thanks.

rminnich:
Alvaro wants to kill the vendoring of this library, which means this is a waste of time.

PR #238, opened 2016-07-31
remove libz.

rminnich:
It was unloved and unwanted, and buggy.


fuchicar:
Libz definitely was not used for anything. The system compiles and starts ok. LGTM

PR #239, opened 2016-07-31
clang: add more compiler names that need -mno-implicit-float

rminnich:
This correctly invoked the version compilers (e.g. clang-3.7) with
the option, but it only seemed to for no-implicit-float with
CC=clang

I have no idea but I hope mkx will allow selection of these
flags for different compilers.


elbing:
It's not as bad as you're thinking. We all things like this many times at work and you know it.
LGTM

PR #244, opened 2016-08-06
ACPI: remove ACPICA source from the kernel.

rminnich:
It is proving to break some principles that we have to not break.

o it is 50KLOC of impenetrable, badly designed, badly written code
o they are asking you to run a giant blob (the ACPI ML) which is
  known to be a source of exploits.
o It's a pain to program with.

All hail Plan 9: taking it out required changing 4 files. That's it.
We get to keep the ioapic.c fix as part of the change.

Tested with both clang and gcc on vmware.
Please test. This is ONLY to get us back to status quo ante.
Next step is to pull in the Akaros code that sets up interrupts
using ACPI. That code is small and nice, so not to worry.

We can continue to consider using ACPICA in user mode; in fact,
I suspect we should.


elbing:
All ok, tested terminal/cpu with new ninep.
LGTM

elbing:
LGTM

sevki:
needs rebase

sevki:
needs another rebase

rminnich:
well, here we go ...

PR #246, opened 2016-08-07
realloc: handle zero-length arrays.

rminnich:
We saw a problem in devacpi.c when realloc was called
to create a zero-length array: it returned a nil pointer,
not a pointer to a zero-length array, which confused the caller.

Consider the difference between:
char *s = "";
char *s = nil;

The first is a pointer to a zero-length string, the second
a NULL pointer. We want realloc to implement the first,
rather than the second.

### DIFF
--- a/sys/src/libc/port/reallocarray.c
+++ b/sys/src/libc/port/reallocarray.c
@@ -28,6 +28,22 @@
 void *
 reallocarray(void *optr, size_t nmemb, size_t size)
 {
+	/* If both size or nmemb are 0, and realloc is called with
+	 * 0, it's equivalent to freeing the memory. realloc is even
+	 * allowed to return nil. Hence this function needs to distinguish
+	 * between 'pointer to 0 length array' and 'nil pointer'. In the
+	 * event that nmemb * size is zero: free optr, and malloc(0),
+	 * i.e. return a pointer to a zero-length array.
+	 * It is entirely possible that it will return optr, since the allocators
+	 * tend to like to optimize that case, but the bookkeeping
+	 * will be correct: the allocator will know that optr points to a
+	 * 0-length allocation.
+	 */
+	if (nmemb == 0 || size == 0) {
+		free(optr);
+		return malloc(0);
+	}
+
 	if ((nmemb >= MUL_NO_OVERFLOW || size >= MUL_NO_OVERFLOW) &&
 	    nmemb > 0 && SIZE_MAX / nmemb < size) {
 		return nil;


sys/src/libc/port/reallocarray.c
```diff
@@ -28,6 +28,17 @@
...
 {
+	/* if both size or nmemb are 0, and realloc is called with
+	 * 0, it's equivalent to freeing the memory. realloc is even
+	 * allowed to return nil. Hence this function needs to distinguish
+	 * between 'pointer to 0 length array' and 'nil pointer'. In the
+	 * event that nmemb * size is zero, we resize to 1 byte, which
+	 * is usable as a pointer to a 0-length array, and further usable
+	 * in later realloc calls.
+	 */
+	if (nmemb * size == 0)
```
dancrossnyc: Sadly this is incorrect; consider what could happens if this overflows.

I'd test explicitly for either zero ('if (nmemb == 0 || size == 0)'), but I think this could be wrapped into the conditional below.

sys/src/libc/port/reallocarray.c
```diff
@@ -28,6 +28,17 @@
...
+		return realloc(optr, 1);
+
 	if ((nmemb >= MUL_NO_OVERFLOW || size >= MUL_NO_OVERFLOW) &&
 	    nmemb > 0 && SIZE_MAX / nmemb < size) {
```
dancrossnyc: One could change this conditional to:

        if (size > 0 && nmemb > 0 &&
            (nmemb >= MUL_NO_OVERFLOW || size >= MUL_NO_OVERFLOW) &&
            SIZE_MAX / nmemb < size)
                return nil;

sys/src/libc/port/reallocarray.c
```diff
@@ -28,6 +28,17 @@
...
+	/* if both size or nmemb are 0, and realloc is called with
+	 * 0, it's equivalent to freeing the memory. realloc is even
+	 * allowed to return nil. Hence this function needs to distinguish
+	 * between 'pointer to 0 length array' and 'nil pointer'. In the
+	 * event that nmemb * size is zero, we resize to 1 byte, which
+	 * is usable as a pointer to a 0-length array, and further usable
+	 * in later realloc calls.
+	 */
+	if (nmemb * size == 0)
+		return realloc(optr, 1);
```
dancrossnyc: I think you can, 'realloc(optr, 0)' here. What we want is whatever realloc malloc() does when passed zero. It should be safe to realloc that.

Then again, I'm not sure what malloc(0) returns. At one point, Dennis Ritchie and Rob Pike (I think) lobbied the ANSI committee to make that return a well-defined, non-nil pointer value, but that never happened.

sys/src/libc/port/reallocarray.c
dancrossnyc: Actually, the more I think about it, the more I don't see why that is necessary. The existing code should fall-through to realloc 'realloc()' if either nmemb or size is zero. Looking at our realloc(), it appears that we explicitly catch the 'size == 0' case and return nil. This is technically perfectly legal; according to the C11 standard, section 7.22.3 para 1, "If the size of the space requested is zero, the behavior is implementation-defined: either a null pointer
is returned, or the behavior is as if the size were some nonzero value, except that the
returned pointer shall not be used to access an object." But looking at 'malloc()', it appears that malloc(0) will probably return a real pointer.

Perhaps the fix is to capture the 'size == 0 || nmemb == 0' case in reallocarray() and 'free(optr); return malloc(0);', which should return something non-nil.  E.g.,

        if (size == 0 || nmemb == 0) {
                free(optr);
                return malloc(0);
        }

sys/src/libc/port/reallocarray.c
dancrossnyc: (I was wrong. I think one should put the following here):

        if (size == 0 || nmemb == 0) {
                free(optr);
                return malloc(0);
        }

dancrossnyc:
Uh oh language lawyer comments ahead:

The thing about:

    Consider the difference between:
    char *s = "";
    char *s = nil;

In the case of 'char *s = "";', 's' actually points to an array of char of length 1, as the '\0' is implicit. Perhaps what we want is something closer to 'char t[0]; char *s = t;'. Unqualified 't' decays to a pointer to a zero-length array object, and 's' is a pointer that object.

The OpenBSD man page for 'reallocarray()' includes this:

'If size or nmemb is equal to 0, a unique pointer to an access protected,
zero sized object is returned.  Access via this pointer will generate a
SIGSEGV exception.'

I think we should return the result of 'malloc(0);' to account for Harvey's 'realloc(p, 0);' returning nil.

sys/src/libc/port/reallocarray.c
```diff
@@ -28,6 +28,22 @@
...
+	 * allowed to return nil. Hence this function needs to distinguish
+	 * between 'pointer to 0 length array' and 'nil pointer'. In the
+	 * event that nmemb * size is zero: free optr, and malloc(0),
+	 * i.e. return a pointer to a zero-length array.
+	 * It is entirely possible that it will return optr, since the allocators
+	 * tend to like to optimize that case, but the bookkeeping
+	 * will be correct: the allocator will know that optr points to a
+	 * 0-length allocation.
+	 */
+	if (nmemb * size == 0){
```
dancrossnyc: Use '||' here to avoid the potential of overflow:

        if (nmemb == 0 || size == 0) { ...

sys/src/libc/port/reallocarray.c
```diff
@@ -28,6 +28,22 @@
 void *
 reallocarray(void *optr, size_t nmemb, size_t size)
 {
+	/* If both size or nmemb are 0, and realloc is called with
```
dancrossnyc: I'd shorten the comment:

    /* malloc(0) returns a valid non-nil pointer, while realloc(p, 0) is equivalent to free(p). */

dancrossnyc:
It looks good to me, but please do use a logical OR instead of multiplication in the test. I'd shorten the comment, but wordsmith it as you see fit.

PR #248, opened 2016-08-11
Updating code to be able being built by gcc 6

elbing:
Added -Wno-frame-address to avoid errors about
__builtin_frame_address in:
- libc/port/getcallerpc.c
- libc/port/malloc.c

And related in kernel and cmd.
Unfortunately, clang doesn't recognize this warning supress option,
instead it has -Wno-address. Suggestions for tweaking old build and new
build tools are welcome.

NOTE: if/else statements indentation is now a warning in case
{} would be missing.

### DIFF
--- a/amd64/include/cflags.json
+++ b/amd64/include/cflags.json
@@ -18,6 +18,9 @@
 			],
 			"/opt/intel/bin/icc": [
 				"-Wno-main"
+			],
+			"/usr/bin/gcc": [
+				"-Wno-frame-address"
 			]
 		},
 		"Cflags": [
--- a/amd64/include/klib.json
+++ b/amd64/include/klib.json
@@ -3,6 +3,9 @@
 		"ToolOpts": {
 			"/usr/bin/clang": [
 				"-mno-implicit-float"
+			],
+			"/usr/bin/gcc": [
+				"-Wno-frame-address"
 			]
 		},
 		"Cflags": [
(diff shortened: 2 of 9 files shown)


elbing:
Since __builtin_frame_address is used across the code, I suggest including the warning supress option by default in old build tool code and some kind of arrangement in new build tool to check what version of gcc is used.

elbing:
Tested, built and booted in kvm, terminal and cpu server.

elbing:
Seeing travis, clang doesn't pass the check :-(.

amd64/include/cflags.json
```diff
@@ -34,7 +34,8 @@
 			"-Wall",
 			"-Wno-missing-braces",
 			"-Wno-parentheses",
-			"-Wno-unknown-pragmas"
+			"-Wno-unknown-pragmas",
+			"-Wno-frame-address"
```
rminnich: If this is gcc-specific it needs to be in toolopts, not here.

elbing:
Removed from sys/src/FLAGS for new build tool, I don't know if we have options like toolopts for mkx.

sevki:
LGTM

sevki:
> NOTE: if/else statements indentation is now a warning in case
{} would be missing. 

I'm liking this a lot. 👍 
missing {} is a pet peeve of mine

PR #252, opened 2016-08-12
reallocarray: if we are asked to allocate 0, return malloc(1)

rminnich:
This ensures that, now and into the future, reallocarracy
of 0 members or zero size will return a non-nil pointer.

We either do this or change the meaning or use of reallocarray.
Currently, the slice code requires a non-nil return.


elbing:
You rule. @dancrossync?

sevki:
LGTM

sys/src/libc/port/reallocarray.c
```diff
@@ -41,7 +41,7 @@ reallocarray(void *optr, size_t nmemb, size_t size)
 	 */
 	if (nmemb == 0 || size == 0) {
 		free(optr);
-		return malloc(0);
+		return malloc(1);
```
dancrossnyc: Malloc(0) already returns a valid pointer. What was the motivation for this change?

rminnich:
it was not returning a valid pointer. It was returning NULL in the kernel, causing slice code to panic.

dancrossnyc:
oh bummer. ok then. :-(

PR #254, opened 2016-08-13
clang-format: always break after function return type.

rminnich:
How's this look? Good? Bad?


.clang-format
```diff
@@ -13,6 +13,7 @@ AllowShortIfStatementsOnASingleLine: false
 AllowShortLoopsOnASingleLine: false
 AllowShortFunctionsOnASingleLine: All
 AlwaysBreakAfterDefinitionReturnType: true
+AlwaysBreakAfterDefinitionReturnType: true
```
sevki: isn't this the same line copied above?

rminnich:
uh, weird, how'd I not see that.

PR #255, opened 2016-08-13
clang-format: turn off sort include and left align pointers

sevki:
it would be wise to use this with `./util/clang-format-diff.py` instead of a big bang format all or a editors format on save function.


rminnich:
LGTM

I don't understand: "it would be wise to use this with ./util/clang-format-diff.py instead of a big bang format all or a editors format on save function."

sevki:
that script will only format lines that you've touched, if you use clang-format in emacs with an on save hook it will format the entire file breaking a bunch of git blame data. same goes with formatting all the files in the project

rminnich:
could we set this up in a pre-commit hoook?

sevki:
yup. I'll update bootstrap to modify pre-commit hook

dancrossnyc:
Forgive me if I get this wrong, I'm tapping this out from my phone at my
daughter's swim lesson.... But left aligned pointers are the devil;
consider:

char* foo, bar;

Visually, it looks like both foo and bar are char pointers, but they are
not: foo is a pointer, bar is a char.

Attaching the '*' to the variable makes this unambiguous.

On Aug 13, 2016 10:13 AM, "Sevki" <notifications@github.com> wrote:

yup. I'll update bootstrap to modify pre-commit hook

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/Harvey-OS/harvey/pull/255#issuecomment-239622835>, or mute
the thread

sevki:
I didn't know that was the behaviour, I even wrote something to verify but you are absolutely right. 

I guess having it left-align in return type makes sense because there are no multiple returns in c. So I guess it's either we put up with it, or patch clang-format to do the right thing, or right a wrapper to it in go to fix the behaviour.

PR #253, opened 2016-08-12
Revert "Revert "ACPI: remove ACPICA source from the kernel.""

rminnich:
This is working for me now. Complete rebuild and boot.

This reverts commit d056e163f082406ad50c323fa6fdcba435d7137c.

PR #257, opened 2016-08-15
ACPI: update the build files to work with ACPI

sevki:
now compiles and is in sync with #253


sevki:
dan said it was ok 😃  
LGTM

PR #258, opened 2016-08-16
UFS: fix which ufs we use.

rminnich:
It's time to move to the one in Harvey-OS.

This is tested and working for supporting harvey guests.

I have a feeling I have the commands not quite right, but OTOH
they work.


bootstrap.sh
```diff
@@ -4,7 +4,8 @@ git submodule init
 git submodule update
 echo Building the build tool...
 GOBIN="$(pwd)/util" GOPATH="$(pwd)/util/third_party:$(pwd)/util" go get -d harvey/cmd/... # should really vendor these bits
-GOBIN="$(pwd)/util" GOPATH="$(pwd)/util/third_party:$(pwd)/util" go install github.com/rminnich/ninep/srv/examples/ufs harvey/cmd/...
+GOBIN="$(pwd)/util" GOPATH="$(pwd)/util/third_party:$(pwd)/util" go get github.com/Harvey-OS/ninep/cmd/ufs harvey/cmd/...
+GOBIN="$(pwd)/util" GOPATH="$(pwd)/util/third_party:$(pwd)/util" go install github.com/Harvey-OS/ninep/cmd/ufs harvey/cmd/...
```
sevki: when you do go get w/o `-d` do you still need to do `go install`

sevki:
this doesn't change the [git submodule](https://github.com/Harvey-OS/harvey/blob/master/.gitmodules) nor the [reference to it's commit ](https://github.com/Harvey-OS/harvey/tree/master/util/third_party/src/github.com/rminnich). either both should be dropped or changed. 

Word of warning: removing git submodules is not as easy as deleting those files.

elbing:
LGTM

elbing:
Removing submodule in another PR

elbing:
What happens with this?

rminnich:
I could not tell if sevki thought it was ok, I'd like his LGTM.

sevki:
LGTM if you apply this patch https://github.com/rminnich/harvey/pull/3

rminnich:
I don't like calling sevki's tool 'nuke' but we can fix that later.

PR #259, opened 2016-08-16
build: harvey_binary type

sevki:
this is going to look like a big commit but it isn't. All it does
is move binary configurations for builds to sys/src/harvey.BUILD
where it defines a harvey_binary macro.

the macro should be used for all std executables by

	load("//sys/src/harvey.BUILD", "harvey_build")

	harvey_binary(
		name="echo",
		srcs=[
			"echo.c"
		],
	)

this PR also aims to clean up the strip commands in `sys/src/9/amd64/BUILD` by moving the option into the `cc_binary` rule and stripping by default for `harvey_binary` macro


elbing:
This made the kernel boot until ipconfig, isn't it? lgtm?

sevki:
yup.

rminnich:
do you still need the old build tool to build the rest of the tree?

sevki:
nope, don't need it

sevki:
@elbing if you want to LGTM it's fine

sevki:
image

sevki:
image

sevki:
can someone LGTM this?

sys/src/cmd/BUILD
```diff
@@ -1,195 +1,86 @@
...
 
-# Order of libs
-# "/$ARCH/lib/libString.a",
-# "/$ARCH/lib/lib9p.a",
-# "/$ARCH/lib/libndb.a",
-# "/$ARCH/lib/libauth.a",
-# "/$ARCH/lib/libbio.a",
-# "/$ARCH/lib/libip.a",
-# "/$ARCH/lib/libc.a"
+CMD_SRCS = [
```
rminnich: Why not just put them all in here?

sys/src/cmd/BUILD
```diff
@@ -1,195 +1,86 @@
...
-    ]
+[harvey_binary(
+	name=c[:-2],
+	srcs=[c],
+) for c in CMD_SRCS]
 
-group(
-    name="cmds",
-    deps=CMDS
+harvey_binary(
```
rminnich: If we make the list above complete, why do we need these?

rminnich:
LGTM
it's a start!

sevki:
@rminnich keeping everything the same and focusing just on changing the target types

PR #260, opened 2016-08-18
QUIET: quiet things down

rminnich:
add kmprint, which is like print but only goes to the
kmesg buffer, i.e. does not pollute the console.

Use kmprint and just reduce the amount of junk printed
so people can see their boot messages.


elbing:
LGTM

[Approved with PullApprove](https://pullapprove.com/Harvey-OS/harvey/pull-request/260/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=Harvey-OS/harvey)

PR #261, opened 2016-08-19
format: fix longstanding formatting bug

rminnich:
For a long time, this code sequence:
	unsigned long x = 0xfeefd000deadbeef;
	seprint(buf, buf+sizeof(buf), "%.2X", x);
	print("%s\n", buf);

would result in this
-21524111

which is not at all what we want.

With this fix, they now result in this:
DEADBEEF

This change includes a simple test for testing formatting, which
we should grow over time: fmt.c


elbing:
LGTM

PR #262, opened 2016-08-20
ACPI: user mode tool to manage IRQs

rminnich:
It's going to work something like this:
when the kernel has an IRQ it can not compute, i.e. needs to run AML, 
it will save information about it in devacpi.

devapic will export a new file, called acpiirq or somesuch. When programs
read that file, it will provide info like this:
t b d f line pin
i.e. the tbdf (broken out) and line and pin. This will avoid any need of
the program to do pci config space ops to find things out.

The program, given this info, will find out the correct ioapic and pin
on ioapic, and write this info back.
t b d f ioapic pin

Note the format of the read and write is the same, just the info is different.

### DIFF
--- a/amd64/include/mach_acpi.h
+++ b/amd64/include/mach_acpi.h
@@ -45,4 +45,5 @@ typedef int8_t INT8;
 typedef int ACPI_THREAD_ID;
 
 #define ACPI_MACHINE_WIDTH 64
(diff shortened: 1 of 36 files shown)


rminnich:
Amazing. It lives.

elbing:
"It's aliiive!" (Victor von Frankenstein).
LGTM

PR #264, opened 2016-08-21
Ramdisk tests all pass (yet code is borken)

keedon:
But only beacuse of a cheat making the minimum alloc size
32k, ramwrite is still broken


sys/src/9/amd64/qmalloc.c
```diff
@@ -565,8 +563,6 @@ smalloc(uint32_t size)
 
 	while((v = malloc(size)) == nil)
```
rminnich: why not just use mallocz?

sys/src/9/port/devramfs.c
```diff
@@ -145,6 +194,10 @@ ramopen(Chan *c, int omode)
 		nexterror();
 	}
 	Chan* ret = devopen(c, omode, nil, 0, ramgen);
+	struct RamFile* file = (struct RamFile*)c->qid.path;
```
rminnich: To me this looks like a panic. In what situation is this not a serious bug or memory corruption?

sys/src/9/port/devramfs.c
```diff
@@ -168,17 +224,24 @@ delete(struct RamFile* file)
...
 }
 
 static void
 ramclose(Chan* c)
 {
         struct RamFile* file = (struct RamFile *)c->qid.path;
+	if (file->magic != RAM_MAGIC)
+		error(INVALID_FILE);
+	qlock(&ramlock);
+	file->busy--;
```
rminnich: just use adec and you can skip the lock.

sys/src/9/port/devramfs.c
```diff
@@ -208,19 +273,44 @@ ramread(Chan *c, void *buf, int32_t n, int64_t off)
...
 
+typedef double Align;
+typedef union Header Header;
+
+union Header {
+        struct {
+                Header* next;
+                uint    size;
+        } s;
+        Align   al;
```
rminnich: what's this for again? To force the allocation alignment? There's better ways to do this, but most allocators currently (or should) align to 64 bytes anyway?

sys/src/9/port/devramfs.c
```diff
@@ -208,19 +273,44 @@ ramread(Chan *c, void *buf, int32_t n, int64_t off)
...
+				qunlock(&ramlock);
+				return 0;
+			}
+			file->data = newfile;
+			file->alloclength = alloclength;
 		}
-		file->data = newfile;
 		file->length = n+off;
 	}
+//	p = (Header*)file->data - 1;
```
rminnich: suggestion. Rather than comment the code out, control it with it
if (ramfsdebug)

why: the test is cheap, and then while debugging we can turn these off and on in gdb, which is handy. Russ pointed out to me once how much he dislikes compiling out debug code, and I tend to agree. Also, bracketing with an if makes sure this will always compile. If you really hate the variable, maybe bracket it with if (0) and it gets snipped out by good compilers (even ken c would do that) but you're still sure it compiles.

sys/src/9/port/devramfs.c
```diff
@@ -253,15 +350,14 @@ ramcreate(Chan* c, char *name, int omode, int perm)
...
-	qunlock(&ramlock);
-
 	mkqid(&c->qid, (uintptr_t)file, 0, file->perm & DMDIR ? QTDIR : 0);
 
-        poperror();
-
         c->offset = 0;
         c->mode = omode;
         c->flag |= COPEN;
+	qunlock(&ramlock);
```
rminnich: we can look later at whether holding this lock for this big chunk impacts performance.

rminnich:
LGTM

might as well get this in, see my suggestions, that way others can test too.

sys/src/9/port/devramfs.c
keedon: This is copied from qmalloc.c so I could get the header. I'll move the qmalloc one into a suitable header file.

sys/src/9/port/devramfs.c
keedon: Will do

PR #267, opened 2016-08-25
Tar is broken

elbing:
This doesn't solve it, but makes it prints real error message.
Inherited from %u modifier:

Before:
tar: bad archive header checksum in /fd/0: name ./...; expected 3o got 3932o

Now:
tar: bad archive header checksum in /fd/0: name ./...; expected 03 got 07534

Guessing %o is now unsigned by default.


elbing:
Tested with local dirs, not only with ramfs.

sevki:
you need to sign-off the commits now

sevki:
yup. o is unsigned octal no need for there

LGTM

PR #268, opened 2016-08-25
sys/include/libc.h: remove a name

rminnich:
We don't put names in .h files. If we did, we'd have a lot
of names in here.


elbing:
LGTM

rminnich:
The names also made no sense in libc, Alvaro told me we should remove them.

PR #55, opened 2016-02-22
ufs fix

sevki:
fixes #54

PR #273, opened 2016-08-28
Fixing detection of Athlon II X2 B24

elbing:
I guess, like my X4, this AMD series must be treated as a K8
Opteron, leaving Phenom (hybrid) as a rare case of K10.

### DIFF
--- a/sys/src/9/amd64/archamd64.c
+++ b/sys/src/9/amd64/archamd64.c
@@ -30,8 +30,6 @@ cpuidinit(void)
 		return 0;
 	machp()->CPU.ncpuinfos++;
 
-	if(memcmp(&machp()->CPU.cpuinfo[0][1], "GenuntelineI", 12) == 0)
-		machp()->CPU.isintelcpu = 1;
 	cpuid(1, 0, machp()->CPU.cpuinfo[1]);
 
 	/*
@@ -67,16 +65,45 @@ cpuidinfo(uint32_t eax, uint32_t ecx, uint32_t info[4])
 	return 1;
 }
 
+static char *
+cpuidname(uint32_t *info0)
+{
+	char *vendorid;
+
+	if(memcmp(&info0[1], "GenuntelineI", 12) == 0) {
+		vendorid = "GenuineIntel";
+		return vendorid;
+	}
+	else if(memcmp(&info0[1], "AuthcAMDenti", 12) == 0) {
+		vendorid = "AuthenticAMD";
+		return vendorid;
+	}
+	else {
+		/* weird hypervisor?, let's pass what it has */
+		vendorid = (char *)&info0[1];
+		print("Warning: CPU unknown, %s\n", vendorid);
+		return vendorid;
+	}
+	return vendorid;
+
+
+}
+
 static int64_t
 cpuidhz(uint32_t *info0, uint32_t *info1)
 {
 	int f, r;
 	int64_t hz;
 	uint64_t msr;
+	char *vendorid;
 
-	//print("CPUID Vendor: %s\n", (char *)&info0[1]);
-	//print("CPUID Signature: %d\n", info1[0]);
-	if(memcmp(&info0[1], "GenuntelineI", 12) == 0){
+	vendorid = cpuidname(info0);
+
+	DBG("CPUID Vendor: %s\n", (char *)&info0[1]);
+	DBG("vendorid: %s\n", vendorid);
+	DBG("CPUID Signature: %d\n", info1[0]);
+
+	if(strcmp("GenuineIntel", vendorid) == 0) {
 		switch(info1[0] & 0x0fff3ff0){
 		default:
 			return 0;
@@ -182,29 +209,26 @@ cpuidhz(uint32_t *info0, uint32_t *info1)
 		}
 		DBG("cpuidhz: 0x2a: %#llx hz %lld\n", rdmsr(0x2a), hz);
 	}
-	else if(memcmp(&info0[1], "AuthcAMDenti", 12) == 0){
+	else if(strcmp("AuthenticAMD",vendorid) == 0){
 		switch(info1[0] & 0x0fff0ff0){
 		default:
 			return 0;
 		case 0x00050ff0:		/* K8 Athlon Venice 64 / Qemu64 */
 		case 0x00020fc0:		/* K8 Athlon Lima 64 */
 		case 0x00000f50:		/* K8 Opteron 2xxx */
+		case 0x00100f60:		/* K8 Athlon II X2 */
 			msr = rdmsr(0xc0010042);
 			r = (msr>>16) & 0x3f;
 			hz = 200000000ULL*(4 * 2 + r)/2;
 			break;
-		case 0x00100f60:		/* K8 Athlon II */
-		case 0x00100f40:		/* Phenom II X2 */
+		case 0x00100f40:		/* Phenom II X2 && Athlon II X4 559 Processor */
 		case 0x00100f20:		/* Phenom II X4 */
 		case 0x00100fa0:		/* Phenom II X6 */
-			msr = rdmsr(0xc0010042);
-			r = msr & 0x1f;
-			hz = ((r+0x10)*100000000ll)/(1<<(msr>>6 & 0x07));
-			break;
 		case 0x00100f90:		/* K10 Opteron 61xx */
 		case 0x00600f00:		/* K10 Opteron 62xx */
 		case 0x00600f10:		/* K10 Opteron 6272, FX 6xxx/4xxx */
 		case 0x00600f20:		/* K10 Opteron 63xx, FX 3xxx/8xxx/9xxx */
+		case 0x00700f00:		/* Athlon II X4 5xxx */
 			msr = rdmsr(0xc0010064);
 			r = msr & 0x1f;
 			hz = ((r+0x10)*100000000ll)/(1<<(msr>>6 & 0x07));


dmgolubovsky:
I tried this to boot with -cpu host (this is the closest I can get to the real CPU), but unfortunately it does not help.
I get:

warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 0]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 1]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 2]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 3]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 4]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 5]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 6]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 7]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 8]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 9]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 12]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 13]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 14]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 15]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 16]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 17]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 23]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 24]

Harvey
die:dumpregs

Anything that I can help with? Can I run something on Linux that gives you the necessary information?

Thanks.

PS I think the problem was in the typo with the word AuthenticAMD (see the attached file)

[amd.diff.txt](https://github.com/Harvey-OS/harvey/files/441625/amd.diff.txt)

Once I fixed the typo I was able to boot with -cpu host.

Hope this helps.

elbing:
That was so because it failed getting the string. But perhaps in past it was wrong in another part. Ok, I'll review that part. Thanks!

elbing:
Seems that strings are ok, but a bit outdated (both, Intel and AMD). So I'll add a check for newer strings from cpuid. GenuineIntel and AuthenticAMD. Maybe one day, vendors decided "why don't we put the right name in the cpus?", who knows...

elbing:
Please test updated patch to see if it works for you. There are still a bunch of CPU out there which could have the old string. Now Harvey checks both.
Thanks!

sys/src/9/amd64/archamd64.c
```diff
@@ -182,29 +183,27 @@ cpuidhz(uint32_t *info0, uint32_t *info1)
 		}
 		DBG("cpuidhz: 0x2a: %#llx hz %lld\n", rdmsr(0x2a), hz);
 	}
-	else if(memcmp(&info0[1], "AuthcAMDenti", 12) == 0){
+	else if((memcmp(&info0[1], "AuthcAMDenti", 12) == 0) ||
```
rminnich: I detect confusion. 
The string is strange because of how they plan 9 cpuid works, not for any good reason. We can talk about it more but basically only ONE of these strings should be here.

sys/src/9/amd64/archamd64.c
```diff
@@ -76,7 +76,8 @@ cpuidhz(uint32_t *info0, uint32_t *info1)
 
 	//print("CPUID Vendor: %s\n", (char *)&info0[1]);
 	//print("CPUID Signature: %d\n", info1[0]);
-	if(memcmp(&info0[1], "GenuntelineI", 12) == 0){
+	if((memcmp(&info0[1], "GenuntelineI", 12) == 0) ||
+		(memcmp(&info0[1], "GenuineIntel", 12) == 0)) {
```
rminnich: This is not really right. You need to go back and see how the cpuid is turning ebx, ecx, and edx into info. The plan 9 usage mean that the info[] array was not a good fit to the string, and we ended up with this weird ordering. 

IOW, you should only need one of these strings, not both. This code is wrong.

rminnich:
don't submit until you resolve the questions around the memcmp.

elbing:
Ok, so only the second one. I'll fix it later.

elbing:
Another round. Cleaner.

rminnich:
LGTM

dmgolubovsky:
Strange, I cannot boot with this file. Was the previous commit (d94d19f291ce735810c2d01e23520d9adc919e99) kept somewhere? Where I changed the CPU string and it worked for me?

Thanks

rminnich:
could you get us the output so we can see what's going on

dmgolubovsky:
Actually Github found the file/commit (it is referenced from my message). Will try it again and post the output.

elbing:
Humm the other way would be reordering string, but seems @dmgolubovsky's cpu has the right registers order. Weird. Just look around internet. I could do an check and force the string with an if in AMD part of the func. Intel surely follows its own rules with ebx edx ecx without exception as is documented.

elbing:
My doubt is:
Should I restore ebx-ecx-edx and reorder the string, being that 100% of intel cpus are ebx-edx-ecx and around 99% of AMD?,
or leave them just now it is, reminding it for other archs (RISCV and aarch64) what should implement its own cpuid asm func, and check if with vendor order ebx-edx-ecx the string is bad and force it with and "if" to put it in line with the rest of the world?.
Also, I could do the check if/force only for the CPUID 100F63 for this kind of Athlon II X2 B24 in line "case 100f60:" of cpuidhz().
What do you think?

elbing:
Or restore cpuidamd64.S and reorder both strings before the memcmp's.  The precedent suggestion has the advantage of we could know what family of processors has what order of registers for cpuid.

elbing:
As I read in:
https://en.wikipedia.org/wiki/CPUID

The order is right now:

	movl	%ebx,0(%rsp)
	movl	%edx,4(%rsp)
	movl	%ecx,8(%rsp)

It's weird that a cpu x86 type has ebx-ecx-edx

Apple (BSD) just reorders the array after cpuid execution:
http://osxbook.com/book/bonus/misc/cpuinfo_x86/cpuinfo_x86.c

    /* Get vendor */
    do_cpuid(0, cpuid_reg);
    bcopy((char *)&cpuid_reg[ebx], &info_p->cpuid_vendor[0], 4); /* ug */
    bcopy((char *)&cpuid_reg[ecx], &info_p->cpuid_vendor[8], 4);
    bcopy((char *)&cpuid_reg[edx], &info_p->cpuid_vendor[4], 4);

Another parts:
http://www.hugi.scene.org/online/coding/hugi%2016%20-%20corawhd4.htm

 ;Get VendorID

         mov eax,0
         CPUID                     ;CPUID!
         mov [edi+0],eax           ;max standard level supported
         add edi,4
         mov [edi+0],ebx
         mov [edi+4],edx
         mov [edi+8],ecx
         mov [edi+12],byte 0       ;vendor ID string+padding

         add edi,16

I can't believe you could have a cpu with different reg order.

elbing:
I think your kvm/qemu version is playing with us.

dmgolubovsky:
I applied this fix on top of my v9p (virtio drivers in progress) branch. Still cannot boot. But I uncommented the debug print in archamd64.c and see this:

With kvm -cpu host:
Harvey
CPUID Vendor: AuthenticAMD�_�����ǯ������_���������
CPUID Signature: 0x00100f62
die:dumpregs

which after masking should match 
0x00100f60:                /* K8 Athlon II X2 */

If I make the CPU string comparison fail (like it was before the fix) I get with -cpu host

Harvey
CPUID Vendor: AuthenticAMD�_�����ǯ������_���������
CPUID Signature: 0x00100f62
mmuinit: vmstart 0xfffffffff0000000 vmunused 0xfffffffff0713000 vmunmapped 0xfffffffff0800000 vmend 0xfffffffff4000000
sys->pd 0x108003 0x108023
cpu0: mmu l3 pte 0xfffffffff0106ff8 = 107023
cpu0: mmu l2 pte 0xfffffffff0107ff8 = 108023
cpu0: mmu l1 pte 0xfffffffff0108c00 = e3
cpu0: mmu l1 pte 0xfffffffff0108c00 = e3
base 0xfffffffff0af9000 ptr 0xfffffffff0af9000 nunits 3475201
allocated
 free ok
parsexsdt: tbl fffffffff0afb04c, end 16
ACPI initialized
        acpiinit();
physalloc color=0 base=0x4000000 size=0x7bfe0000
before mpsinit: acpionly 1, maxcores 1024
before mpacpi, maxcores 1024
APIC lapic paddr 0xfee00000, flags 0x00000001
ASlapic 0
apic proc 0/0 apicid 0 
ASioapic 0
ioapic 0 addr fec00000 ibase 0 
apic0: hz 668427820 max 6684278 min 66842 div 3
acpi_make_rdt(0x5ffff00 1 255 0x4)
interrupt 11 -> 13 ... Success
interrupt 10 -> 12 ... Success
interrupt 10 -> 12 ... Success
interrupt 11 -> 13 ... Success
interrupt 11 -> 13 ... Success
interrupt 10 -> 12 ... Success
interrupt 10 -> 12 ... Success
interrupt 11 -> 13 ... Success
bus dev type vid  did intl memory
0   0/0 06 00 00 8086 1237 A   0  
0   1/0 06 01 00 8086 7000 A   0  
0   1/1 01 01 80 8086 7010 A   0  4:0000c2a1 16 
0   1/3 06 80 00 8086 7113 B   9  
0   2/0 03 00 00 1b36 0100 B  10  0:f4000000 67108864 1:f8000000 67108864 2:fc090000 8192 3:0000c201 32 
0   3/0 02 00 00 10ec 8139 B  11  0:0000c001 256 1:fc092000 256 
0   4/0 02 00 00 1af4 1000 B  13  0:0000c221 32 1:fc093000 4096 
0   5/0 00 02 00 1af4 1009 B  12  0:0000c101 64 1:fc094000 4096 
0   6/0 00 02 00 1af4 1009 B  12  0:0000c141 64 1:fc095000 4096 
0   7/0 07 80 00 1af4 1003 B  13  0:0000c241 32 1:fc096000 4096 
0   8/0 01 00 00 1af4 1004 B  13  0:0000c181 64 1:fc097000 4096 
0   9/0 01 00 00 1af4 1001 B  12  0:0000c1c1 64 1:fc098000 4096 
0  10/0 00 ff 00 1af4 1005 B  12  0:0000c261 32 
0  11/0 00 ff 00 1af4 1002 B  13  0:0000c281 32 
acpi_make_rdt(0x5ffff00 12 255 0x30)
acpi_make_rdt(0xc001800 11 0 0xc)
#l0: rtl8139: 100Mbps port 0xc000 irq 11 tu 1514: 525400123456
getpgszi: size 0x200000 not found
panic: cpu0: newseg: getpgszi 2097152
die:wait forever

So it goes a bit further and dies. Perhaps because none of CPU features were processed properly.

Going back to my branch v9p.

With -cpu host: die:dumpregs (right away).
With -cpu pentium: boots.
With -cpu kvm64: boots.

Switched to master, pulled the latest updates.

With -cpu host: die:dumpregs (right away).
With -cpu pentium: boots.
With -cpu kvm64: boots.

But when I uncommented CPU string output in master, I got AuthcAMDenti (BTW it outputs so even with -cpu pentium). CPUID for "pentium" is 0x00000543. I'm afraid that in the KVM mode there is not much done to emulate "other than host" CPU.

That's all I got right now. Hope this helps.

Thanks

elbing:
Ok, I have to restore cpuidamd64.S to the original order of registers. In Harvey, cpuid function takes three arguments and not only gets vendor id, it gets extended features too. Usually this is done in two different functions in other OSes.

In order to avoid the lack of having two almost equal functions, it's better to have cpuid getting 3 args and reorder the vendor id string in C in archamd64.c a la BSD.

Ron was in right, I'll do a cpuidname func which returns vendor id in right format and then passing it to cpuidhz for checkings.

---
My doubt is in archamd64.c:33 (cpuidinit)

```C
if(memcmp(&machp()->CPU.cpuinfo[0][1], "GenuntelineI", 12) == 0)
		machp()->CPU.isintelcpu = 1;
```

I think that check is old and hasn't any sense if later it will check if it's AMD or Intel. I'll move ```CPU.isintelcpu = 1``` to the right part of Intel checks in cpuidhz unless I found in other parts of the kernel source that feature was used (and where and why). In cpuidinit I can't pass an array element directly or pointer to vendor id string because it will be run from machp(). That's why I think it's better putting it out of there.

elbing:
I'll upgrade the branch too.
Thanks @dmgolubovsky .

elbing:
Please, test the last patch with new cpuidname function. If it still gives you and error, I could add another "if" with a tautologic vendorid == "AuthenticAMD" for cases in what hypervisors has a bug in reg order or anything.
--
@rminnich I'm sure you will see many things suitable of optimization here, feel free of crush me unmercifully :-)

sys/src/9/amd64/archamd64.c
```diff
@@ -67,16 +67,43 @@ cpuidinfo(uint32_t eax, uint32_t ecx, uint32_t info[4])
...
+	if(memcmp(&info0[1], "GenuntelineI", 12) == 0) {
+		vendorid = "GenuineIntel";
+		return vendorid;
+	}
+	else if(memcmp(&info0[1], "AuthcAMDenti", 12) == 0) {
+		vendorid = "AuthenticAMD";
+		return vendorid;
+	}
+	else {
+		vendorid = nil;
```
rminnich: Why not just return strdup(&info[1]) here? Just because you do not know it does not mean it's invalid. It's unlikely but, well, you never know. Still, I like the approach.

I advocate the strdup instead of just returning &info[1] b/c we don't know what the caller might do with the passed-in data. It may not matter however.

Overall, for now, this looks fine.

sys/src/9/amd64/archamd64.c
```diff
@@ -67,16 +67,43 @@ cpuidinfo(uint32_t eax, uint32_t ecx, uint32_t info[4])
...
 	int64_t hz;
 	uint64_t msr;
+	char *vendorid;
+
+	vendorid = cpuidname(info0);
 
-	//print("CPUID Vendor: %s\n", (char *)&info0[1]);
-	//print("CPUID Signature: %d\n", info1[0]);
-	if(memcmp(&info0[1], "GenuntelineI", 12) == 0){
+	print("CPUID Vendor: %s\n", (char *)&info0[1]);
```
rminnich: did you want to print the vendorid here, since you will panic if it's not valid anyway?

rminnich: on nvm sorry got careless.

sys/src/9/amd64/archamd64.c
```diff
@@ -30,8 +30,8 @@ cpuidinit(void)
 		return 0;
 	machp()->CPU.ncpuinfos++;
 
-	if(memcmp(&machp()->CPU.cpuinfo[0][1], "GenuntelineI", 12) == 0)
-		machp()->CPU.isintelcpu = 1;
+	//if(memcmp(&machp()->CPU.cpuinfo[0][1], "GenuineIntel", 12) == 0)
```
rminnich: Do you not want to set isintelcpu any more?

sys/src/9/amd64/archamd64.c
elbing: Seems fine, in cpuidhz, if cpu is not known (VIA or Cyrix have a 64 bit processors these days?), fall sinto another post-routine to calculate through clock. So strdup is fine. I'll change that. Panic is more for testing purposes on @dmgolubovsky kvm issue.

rminnich:
If it works for all of you, let's do it. 
I'll let someone else LGTM who can test it :=)

sys/src/9/amd64/archamd64.c
elbing: That are traces. They must be cleaned when WIP in tittle will be gone. Also, the isintelcpu useless check and element for mach struct (not used across all the kernel, I think it's and old jmk's test for debugging or anything) is commented now, but I'll remove it when this PR was working for all of us.

sys/src/9/amd64/archamd64.c
elbing: Exactly that I was commenting in previous line-comment. It's a test, it's not used at any place in kernel. Probably and anciente debugging test or anything. It will be gone when WIP dissapeared in tittle.

elbing:
Let's wait for @dmgolubovsky tests. I'll clean the code and add strdup if all goes well.

dmgolubovsky:
Does not boot with -cpu host:

warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 0]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 1]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 2]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 3]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 4]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 5]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 6]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 7]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 8]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 9]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 12]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 13]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 14]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 15]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 16]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 17]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 23]
warning: host doesn't support requested feature: CPUID.80000001H:EDX [bit 24]

Harvey
CPUID Vendor: AuthcAMDenti�_�����9�������_�����q���
vendorid: AuthenticAMD
CPUID Signature: 1052514
die:dumpregs

Boots with -cpu kvm64 and -cpu pentium

Printing CPUID information I think is a good idea (maybe CPUID signature could be printed in hex).

So, just like before. BTW where do these strings "host doesn't support..." come from? I cannot find them in the sources.

Just in case, my KVM version (under Ubuntu 16.04):
kvm -version
QEMU emulator version 2.5.0 (Debian 1:2.5+dfsg-5ubuntu10.4), Copyright (c) 2003-2008 Fabrice Bellard

sys/src/9/amd64/archamd64.c
```diff
@@ -67,16 +67,43 @@ cpuidinfo(uint32_t eax, uint32_t ecx, uint32_t info[4])
...
+	char *vendorid;
+
+	vendorid = cpuidname(info0);
 
-	//print("CPUID Vendor: %s\n", (char *)&info0[1]);
-	//print("CPUID Signature: %d\n", info1[0]);
-	if(memcmp(&info0[1], "GenuntelineI", 12) == 0){
+	print("CPUID Vendor: %s\n", (char *)&info0[1]);
+	print("vendorid: %s\n", vendorid);
+	print("CPUID Signature: %d\n", info1[0]);
```
rminnich: This should be printed in hex.

sys/src/9/amd64/archamd64.c
```diff
@@ -182,29 +209,26 @@ cpuidhz(uint32_t *info0, uint32_t *info1)
...
 		case 0x00020fc0:		/* K8 Athlon Lima 64 */
 		case 0x00000f50:		/* K8 Opteron 2xxx */
+		case 0x00100f60:		/* K8 Athlon II X2 */
 			msr = rdmsr(0xc0010042);
 			r = (msr>>16) & 0x3f;
 			hz = 200000000ULL*(4 * 2 + r)/2;
 			break;
-		case 0x00100f60:		/* K8 Athlon II */
-		case 0x00100f40:		/* Phenom II X2 */
+		case 0x00100f40:		/* Phenom II X2 && Athlon II X4 559 Processor */
```
rminnich: Looks like you need a case statement for 100fc2 here. But is it 64 bits?

rminnich: ah you don't need 100fc2, as the low nibl is cleared. 100fc0 will do it. But you need to find out what that is.

elbing:
Seems rare, because that CPUID signature 1052514 should be an AMD Athlon(tm) II X2 215 Processor and it falls into 100F62 [see 1052514  here](http://www.insanelymac.com/forum/topic/281450-mavericks-kernel-testing-on-amd-formerly-mountain-lion-kernel-testing-on-amd/page-174) and [Athlon II X2 215 specifications here](http://www.cpu-world.com/cgi-bin/CPUID.pl?CPUID=51522) , but I'll try your suggestion. In comment I'll write "weird kvm".

rminnich:
no you're right I typed it wrong. 
I think we need to see a uname -a for that cpu. There's something weird going on here. 
But it's not a harvey problem at this point I think.

rminnich:
IOW you already cover the case, you don't need more. It looks to me like a problem in the system in question.

elbing:
I think strdup in final ```else``` in cpuidname() will resolv this case definitevely bypassing weird hypervisors issue.

elbing:
Ok, @dmgolubovsky : this should solve the issue. Please, last test.
Thanks for your patience.

elbing:
If cpu is now not recognized, cpuidhz will return 0, then in archhz, it will fall here:

```C
iprint("archhz, cpuidhz failed, going to i8254hz\n");
	return i8254hz(info0, info1);
```

Calculating speed through clock and going forward. I used a warning in cpuidhz which will notify us when we have an standard real cpu working and not recognized through cpuid (remember: not only the vendor, extended features too).

dmgolubovsky:
Same as before. Does not boot with -cpu host, boots with others. It seems that there is a problem in rdmsr:

Harvey
CPUID Vendor: AuthcAMDenti�_�������������_���������
vendorid: AuthenticAMD
CPUID Signature: 100f62
1die:dumpregs

Note "1" before die: it comes from the debug I added:

        else if(strcmp("AuthenticAMD",vendorid) == 0){
                switch(info1[0] & 0x0fff0ff0){
                default:
DBG("0");
                        return 0;
                case 0x00050ff0:                /* K8 Athlon Venice 64 / Qemu64 */
                case 0x00020fc0:                /* K8 Athlon Lima 64 */
                case 0x00000f50:                /* K8 Opteron 2xxx */
                case 0x00100f60:                /* K8 Athlon II X2 */
DBG("1");
                        msr = rdmsr(0xc0010042);
DBG("2");
                        r = (msr>>16) & 0x3f;
                        hz = 200000000ULL*(4 * 2 + r)/2;
                        break;

We see 1 but do not see 2.

uname -a as requested:

uname -a
Linux hpcompaq 4.4.0-36-generic #55-Ubuntu SMP Thu Aug 11 18:01:55 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux

And link to cpuinfo:

https://github.com/Harvey-OS/harvey/files/440495/cpuinfo.amd.txt

Hope this helps.

Thanks.

rminnich:
your environment can't be right.
The cpuid is returning the string for an amd. It's claiming it's type 100f62.Yet you can't read that msr?

OK, next step.
run your qemu -s -S. It will then wait for you to attach w/gdb. put a breakpoint at this function. 
set what to display:
display/i $pc
Run until it is called, and step through the msr function:
stepi

and let's see what you get. Also, boot linux with -cpu host and use the rdmsr command:
rdmsr 0xc0010042

and let us know what that returns. I don't trust your kvm any more :-)

elbing:
Try moving case f60 to just before 100f90. Maybe I'm wrong and that kind of cpu needa rdmsr to 064. And tell me the result. Or you can use gdb as Ron says.

rminnich:
I would really like to see the result of booting linux on that kvm with -cpu host and do the rdmsr command.

dmgolubovsky:
Moving case f60 did not help, dies on rdmsr same way. I tried to run rdmsr on the *host*: it says "file not found", and I see from strace that it tries to read /dev/cpu/0/msr, but there is only /dev/cpu/microcode.

Too old CPU? The computer was made in 2010 tol my knowledge.

Linux (coreOS) boots fine with CPU=host (but there is no /dev/cpu at all, will setup something else to see). But absence of /dev/cpu/0 on the host makes me think that perhaps I just need to use -cpu kvm64...

Thanks.

elbing:
No, I don't think so. I think that kvm with -cpu host has some bug. Any way, that "file not found" could be happening due to a buggy version of msr-tools in linux [(for instance)](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=721825), not the cpu itself. But, it's clear that Harvey recognizes an AMD cpu. I'm pretty sure that only two address can be asked about cpuid, and those what use ending in 42 comes from 2007. K10 and hybrids uses address ending 64. There's no other possibility unless cpu couldn't be recognized as AMD or Intel (third limit case), in whose case it will fall in a clock calculation routine. But **never** could hang the boot. I checked Athlox II X2 and its 100f60 case is in right place (look at NxM sources, for instance), when I implemented cpuidhz a year ago after my ThinkPad was failing due to wrong functions for Intel, maybe I thought Athlon II X2 was like a Phenom, but now it's right. More... I'm pretty sure that PR looks as it must look.
At the begining your kvm -cpu host was being recognized as AMD and hanging, when you changed the vendor string to right order, it didn't recognize the CPU due to ebx-ecx-edx wrong order in our asm, and jumped into archhz clock calculation. When I made the subsequent fixes, CPU was always being recognized at any case (right regs order, cpuidname func...) and, then, hanging all the time since that. The advantage is that now we have a better process to do all of this, but **if your kvm -cpu host is well recognized it hangs**. Look and think about all made until now.
In my ThinkPad L420, if I boot kvm with -cpu Phenom it sees my real Intel. If I boot with -cpu -host it sees my real Intel, no matter what I could do, it always sees an Intel. So, If you use kvm paravirtualizer, don't try to lie to the machine and let it boot with an Intel. Otherwise, don't use kvm. Just qemu alone.
What if you just try kvm without -cpu option? does it fails too? If not, fine. If yes, just use kvm64. Important part is your virtio work and Harvey working. Not resolving possible kvm bugs. I'm saying this because the last year and half, we entered in deep debugging process due to qemu bugs. One of them breaking SMP at early days, for example.

----

I think we could merge this, anyway.

dmgolubovsky:
OK from me to merge. You are right, I cannot pretend it is an Intel when real host is AMD (I mentioned that earlier). Thanks for your efforts.

elbing:
Thanks for your patience. Let's go back to virtio. It will open many doors :-).
And think that at the end, we have better routines to detect cpu's. It wasn't a waste of time.

dmgolubovsky:
Right. Virtio needs to fix the multiple mouse interrupts first, and the interrupt table allocation second. I'm going to look into this tonight (need some time to learn ACPI/APIC stuff - hope to hear from Ron as well). The merge button seems available to me - can I do it, or will you?

elbing:
I asked Ron about last details here. When he told me about, I'll merge this.
Thanks!

rminnich:
typically, with linux, to read msrs, you first do
modprobe msr

and then the rdmsr command. rdmsr has worked on linux for 15 years so I doubt your machine is too old.

rminnich:
LGTM

dmgolubovsky:
OK, modprobe msr helped (I thought this module was loaded)

On the host:

sudo rdmsr 0xc0010042
rdmsr: CPU 0 cannot read MSR 0xc0010042
sudo rdmsr 0xc0010064
800001e13c00140e

This was on the host.

On a guest (Fedora 24 running on kvm -cpu host, cpuinfo shows the same model as host):

cannot copy-paste, but it says "cannot read msr 0xc0010064"

Consistent to some degree. rdmsr works on the host, but not in the guest.

So perhaps a bug in KVM, somehow Linux works around that.

Thanks.

rminnich:
This kind of confirms it: looks like a kvm bug. Thanks for working on this.

rminnich:
linux probably works around it because they have rdmsr_safe, which returns an error, not a panic :-)

I think we need that. Later.

rminnich:
Merged at Alvaro's request.

PR #274, opened 2016-08-29
Generic virtqueue kernel driver (WIP)

dmgolubovsky:
These commits add (incomplete) code for the generic virtqueue kernel driver. The driver is aimed to handle each virtqueue as a pipe, no particular logic for an emulated device is in the driver; it is expected to be implemented in userland.
The reason this is pushed now is recently discovered IRQ conflict between virtio other devices.


sys/src/9/amd64/devvq.c
```diff
@@ -0,0 +1,595 @@
...
+// Map device identifiers to descriptive strings to display as virtio device names.
+
+typedef struct
+{
+	uint16_t did;
+	char *desc;
+} didmap;
+
+static didmap dmtab[] = {
+	PCI_DEVICE_ID_VIRTIO_NET, "virtio-net",
```
rminnich: I'm surprised the compiler did not complain about lacking {} around each struct initializer. 

however ...
since the device ids are (intentionally) a very compact zero-based range I would suggest a change here. 
The struct is a good idea. but
typedef struct 
{
  char *name
} didmap;

static didmap dmtabp[] = {
[PCI_DEVICE_ID_VIRTIO_BLOCK] {.name - "virtio-block"},
and so on
};

Then you don't have to search; just index with the device id.

sys/src/9/amd64/devvq.c
```diff
@@ -0,0 +1,595 @@
...
+		vqinterrupt(q);
+	return 0;
+}
+
+static char *
+mapdev(vqctl *vc)
+{
+	char *dmap = nil;
+	for(int i = 0; i < nelem(dmtab) ; i++) {
+		if(vc->pci->did == dmtab[i].did) {
```
rminnich: If you use the array I suggested above this nicely gets rid of this for loop.

sys/src/9/amd64/devvq.c
rminnich: you can use .name == nil as a test for whether it's a valid entry.

sys/src/9/amd64/devvq.c
```diff
@@ -0,0 +1,595 @@
...
+			vc->ownpid = -1;
+			break;
+		default:
+			break;
+	}
+}
+
+
+
+// Reading from the device configuration area is byte-by-byte (8bit ports) to avoid dealing with
```
rminnich: we can talk about this later but this is going to be a huge perfomance hit. Each vmexit will require emulation on the part of the host, and that won't be cheap. I think this will be a performance issue.

sys/src/9/amd64/devvq.c
```diff
@@ -0,0 +1,595 @@
...
+//print("vqwrite %d %d %d\n", TYPE(c->qid), n, offset);
+	return -1;
+}
+
+// Scan virtqueues for the given device. If the vqs argument is not nil then
+// nvq is expected to contain the length of the array vqs points to. In this case
+// populate the Virtq structures for each virtqueue found. Otherwise just return
+// the number of virtqueues detected. The port argument contains the base port
+// for the device being scanned.
+// Portions of code are borrowed from the 9front virtio drivers.
```
rminnich: OK we can't just take code that is "borrowed". We have to manage the license correctly. Did you talk to cinap about the license question? He's been very nice about releasing his code to us under MIT or BSD license.

sys/src/9/amd64/devvq.c
```diff
@@ -0,0 +1,595 @@
...
+void
+vqinit(void)
+{
+	print("\nvq initializing\n");
+	nvq = findvdevs(nil);
+	print("\nvq: found %d devices\n", nvq);
+	if(nvq == 0) {
+		return;
+	}
+	cvq = malloc(nvq * sizeof(vqctl *));
```
rminnich: Just checking: does this need to be zero'd? if so use mallocz. If there's even a question, use mallocz.

sys/src/9/amd64/virtio_def.h
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * No part of Harvey operating system, including this file, may be copied,
+ * modified, propagated, or distributed except according to the terms
+ * contained in the LICENSE.gpl file.
+ */
+
+// Common definitions for virtio in Harvey. Some definitions are borrowed from
+// other sources, where importing a whole file did not seem feasible.
+
+// Virtio-pci device type constants, from
```
rminnich: You need to name the sources however.

sys/src/9/amd64/virtio_def.h
```diff
@@ -0,0 +1,44 @@
...
+#define PCI_DEVICE_ID_VIRTIO_NET         0x1000
+#define PCI_DEVICE_ID_VIRTIO_BLOCK       0x1001
+#define PCI_DEVICE_ID_VIRTIO_BALLOON     0x1002
+#define PCI_DEVICE_ID_VIRTIO_CONSOLE     0x1003
+#define PCI_DEVICE_ID_VIRTIO_SCSI        0x1004
+#define PCI_DEVICE_ID_VIRTIO_RNG         0x1005
+#define PCI_DEVICE_ID_VIRTIO_9P          0x1009
+
+// Specific definitions for virtio-9p devices, from
+// https://git.kernel.org/cgit/linux/kernel/git/stable/linux-stable.git/tree/include/uapi/linux/virtio_9p.h
```
rminnich: Sorry, I just realized you did a great job of naming the source.

sys/src/9/amd64/virtio_pci.h
```diff
@@ -0,0 +1,201 @@
...
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ */
+
+#ifndef _LINUX_VIRTIO_PCI_H
```
rminnich: it's ok to remove stuff like these include guards, on the next pass. We don't do them in harvey.

sys/src/9/amd64/virtio_ring.h
```diff
@@ -112,13 +112,31 @@ struct virtq_used {
         /* Only if VIRTIO_F_EVENT_IDX: le16 avail_event; */
 };
 
-struct virtq {
-        unsigned int num;
+/* borrowed from 9front sdvirtio: a structure to sleep/wake a process waiting on I/O */
+
+typedef struct rock {
```
rminnich: let's make sure they are ok with that, please.

rminnich:
This is wonderful. I will
-1
until we make sure we're all clear on the 9front imports. What license was the code you imported? It can't be Lucent.

elbing:
Just mark in title WIP and it won't be merged.
Cinap's code MIT. Mischief's one GPL. They talked to me so.

rminnich:
how we import code:
1. import the whole file(s),unchanged, and commit it. Make sure there is a license statement in the file(s) (if there is not one, add it from the project you got it from), and that it is GPL compatible, meaning: BSD, or MIT. It must not be Apache-[2,3] or Lucent Public Licnse. It doesn't need to compile. 

2. copy the the file(s) to the file name(s) you want. commit that. It doesn't need to compile. 
Change the files you want. If you don't want to keep the 9front file, remove it. It doesn't need to compile.

3. Get your code to compile. 

This way, it's easy to go back and show anyone where things came from. We want other projects to be comfortable that we are following the rules.

rminnich:
How's this one coming? It would be really nice to have.

rc/bin/termrc
```diff
@@ -9,7 +9,7 @@ terminal = "ninep"
 mntgen -s slashn && chmod 666 /srv/slashn
 
 # bind all likely devices (#S was bound in boot)
-for(i in f t m v L P u U '$' Σ κ)
+for(i in f t m v L P Q u U '$' Σ κ 'α')
```
rminnich: what is alpha? Can we change it to 'Z' if it's supposed to be acpi?

sys/src/9/386/pci.c
```diff
@@ -401,6 +401,22 @@ pcicfginit(void)
 				/* reset the cardbus */
 				pcicfgw16(p, PciBCR, 0x40 | pcicfgr16(p, PciBCR));
 				delay(50);
+
```
rminnich: let's fix this another way, definitely we don't want to do this.

sys/src/9/amd64/devvq.c
rminnich: This PR is based on another PR, or standalone, or ... what's going on here? Sorry if I am missing something.

rminnich:
it's true. I really am dumber without coffee. Sorry for the confused review. If you want I'll try to get this code in in the manner I laid out -- clean commit of unchanged stuff, etc. Let me know if I can help.

dmgolubovsky:
It is going as time permits (not much of it I have, but I keep looking into this). I am trying to decide how I integrate the 9p-virtio pipe into the existing infrastructure of networked mounts. An early thought (did not even try to code this) to have virtio queue/pipe as another type of networked connection, so it could be dialed from the standard libc (using notation like virtio!mount_tag) - because to my understanding it should be just a pipe. Other virtio devices may need different approach, but my primary goal is to have native 9p to the host.

Regarding Z vs Alpha: I just needed to mount APIC into the namespace to inspect the MADT so found which letter it was and added. It was Alpha first, then it became Z.

Per other findings in the code: thanks for looking into it, but it may change in any direction.

dmgolubovsky:
This PR is obsoleted by #308 closing it now.

PR #277, opened 2016-08-30
dist: remove dist/

rminnich:
It was never very good and we don't intend to keep it.


sevki:
LGTM

PR #278, opened 2016-08-30
sys/test: first test script, tests boot

sevki:
this is a first pass at the boot test. 

Signed-off-by: Sevki s@sevki.org


sys/test/BUILD
```diff
@@ -0,0 +1,29 @@
...
+        "dump,file=/tmp/vm0.pcap",
+    ],
+    redir=[
+        "tcp:9999::9",
+        "tcp:17010::17010",
+        "tcp:17013::17013",
+    ],
+    append="service=cpu nobootprompt=rc maxcores=1024 fs=10.0.2.2 auth=10.0.2.2 nvram=/boot/nvram nvrlen=512 nvroff=0 -user=harvey",
+    cmds=[
+        "/boot/ls",
```
sevki: right :)

sevki:
yes sir

sevki:
inside the harvey directory 

```
go get -u sevki.org/build/cmd/build
build -v //sys/test:boot
```

BUILD
```diff
@@ -94,11 +94,15 @@ kernel(
 		"amd64/bin/date": "bin/date",
```
rminnich: What I would prefer, given how smart bldy is, is to just have the kernel depend on //sys/src/cmd:cmd

sys/src/9/amd64/BUILD
```diff
@@ -266,14 +267,23 @@ cc_binary(
 		"-T",
 	]
 )
+move(
+	name="factotum",
+	deps=[
+		"//sys/src/cmd/auth/factotum:factotum",
+	],
+	installs={
+		"bin/auth/factotum": "bin/factotum",
```
rminnich: er, what? Why?

BUILD
rminnich: where are these things being moved? Maybe I am just not understanding what you're doing here. I don't get it.

PR #280, opened 2016-09-03
mouse: don't do mouseenable twice; just in main.c

rminnich:


rminnich:
ps2 users: please test. But this should be fine.

elbing:
I think this does not need to be tested. It would requiere a mobo with PS/2 and the mouse device for that. If anyone in future opens an issue for this, ok. If not, we could wait forever. I got only one PS/2 in my Asus A1M1 but keyboard is plugged there. Since I'm using it as cpu server, I could test in future plugging an usb keyboard and leaving the PS/2 connector for mouse. For now, I haven't an usb keyb.

rminnich:
This lonely commit needs an L G T m

elbing:
LGTM

PR #282, opened 2016-09-06
ACPI changes to get us to a user-mode AML interpreter that can set IRQs.

rminnich:


elbing:
It boots, but I think you will want the output of kmesg. As you already know: Asus A1M1 realhw.

[out.txt](https://github.com/Harvey-OS/harvey/files/458004/out.txt)

If you think it's ok for now, let me know for an lgtm.

elbing:
LGTM

rminnich:
well, let's hope I did not just break a lot of things ...

elbing:
Late, but you wanted to see it:

image

Real ASUS A1M1, just to have constancy about it.

PR #296, opened 2016-09-14
snoopy: get snoopy to build.

rminnich:
This actually works now. We can snoop the network again.

If someone can fix the yerror hack I did I'll be grateful.


rminnich:
to install snoopy, for not, 
build sys/src/cmd/ip/snoopy

elbing:
LGTM

rminnich:
we now have snoopy for harvey, which is going to be very handy.

PR #299, opened 2016-09-14
AHCI: it's 64 bits, not 32 bits

rminnich:
It was never setting the hi 32 bits in any DMA address.
I moved the driver to amd64 from 386 to indicate that
it's really a 64 bit driver now.


elbing:
LGTM

PR #302, opened 2016-09-15
Fix which errstr is used for syscallfmt.

rminnich:
We should use syserrstr, not errstr.


elbing:
LGTM

PR #301, opened 2016-09-14
Adding /mnt/plumb for making happy drawterm

elbing:
Change-Id: I9c6d9dac3e5def4471291361f1179c0f220093cf

PR #303, opened 2016-09-17
Create /mnt/plumb in case it does not exist

dmgolubovsky:
Plumber fails, and rio does not start if it does not exist.

### DIFF
--- a/usr/harvey/lib/profile
+++ b/usr/harvey/lib/profile
@@ -8,6 +8,7 @@ fn cd { builtin cd $* && awd }  # for acme
 
 switch($service){
 case terminal
+	test ! -d /mnt/plumb && mkdir /mnt/plumb
 	plumber
 	echo -n accelerated > '#m/mousectl'
 	echo -n 'res 3' > '#m/mousectl'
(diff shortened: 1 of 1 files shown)


floren:
Don't we have that cool file server that creates whatever directory you
want under /mnt?

hdonnay:
Yeah, I think we should modify the script to use `slashn` on /mnt

elbing:
I thought on this, but if it never was in profile, I guessed it shouldn't be.
If someone tweak slashn or whatever for this, this PR and mine will be abandoned. But think that for now Harvey doesn't boot without any of two. So think quickly.

rminnich:
/mnt is where things go that are more system level. /n is other stuff and where slashn server runs.

The problem with this change to profile is you are assuming the nodes you cpu into will let you do the mkdir, i.e. that /mnt is writeable to you. That's not really a good assumption. In fact it's almost certainly wrong because /mnt is more of a system level thing like /usr or /lib.

You can run a server that lets you write mkdir /mnt/plumb but that's just more overhead to create a directory that should already exist.

Let's just go ahead and make sure /mnt/plumb is in our repo by adding it. 

-2

usr/harvey/lib/profile
```diff
@@ -8,6 +8,7 @@ fn cd { builtin cd $* && awd }  # for acme
 
 switch($service){
 case terminal
+	test ! -d /mnt/plumb && mkdir /mnt/plumb
```
rminnich: Let's keep it simple. /mnt/plumb should be there, so in a later PR I put it there.

elbing:
Fixed in #301

dmgolubovsky:
OK, whatever is the right solution, mine was just a quick fix.

Thanks

elbing:
And always welcome. Thanks to you.

PR #306, opened 2016-09-18
Revert "Bring in the virtqueue definitions from the virtio spec 1.0. Make sure"

rminnich:
This reverts commit db0ccedae86f62178dc57f41b289fd0f38a18bb1.

We need to bring the external code in a bit more staged.

First, bring the files in with the source (kernel) and git ref
of where they're from. Then start changing them.

Conflicts:
	sys/src/9/386/pci.c
	sys/src/9/amd64/virtio_ring.h
	sys/src/9/port/devpci.c

### DIFF
--- a/sys/include/virtio_ring.h
+++ b/sys/include/virtio_ring.h
@@ -1,0 +1,164 @@
+#ifndef VIRTIO_RING_H
+#define VIRTIO_RING_H
+/* An interface for efficient virtio implementation.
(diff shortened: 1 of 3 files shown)


rminnich:
I'm doing an LGTM since the gh review was ok. We'll need to figure out if gh can block merges or not.

rminnich:
LGTM

PR #307, opened 2016-09-18
Remove line 335 in syscall.c which caused unneeded console output

dmgolubovsky:
The line was likely left by mistake.


rminnich:
My bad. Many thanks for fixing this. I'm trying to track down the errstr problem :-)

LGTM

PR #308, opened 2016-09-19
virtio: bring in include files from Linux 2c9b3ebd5913c2d1371b749a8057ac32972b410d (OK to merge as proposed)

rminnich:

### DIFF
--- a/sys/include/virtio_blk.h
+++ b/sys/include/virtio_blk.h
@@ -1,0 +1,149 @@
+#ifndef _LINUX_VIRTIO_BLK_H
+#define _LINUX_VIRTIO_BLK_H
+/* This header is BSD licensed so anyone can use the definitions to implement
+ * compatible drivers/servers.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. Neither the name of IBM nor the names of its contributors
+ *    may be used to endorse or promote products derived from this software
+ *    without specific prior written permission.
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS ``AS IS'' AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED.  IN NO EVENT SHALL IBM OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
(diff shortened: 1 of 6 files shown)


elbing:
wow, does it work?

rminnich:
no, it doesn't, and I need to do some tweaking. But I've got it in a more harvey regular form.
golubovsky did a really nice job on this.

dmgolubovsky:
Ron

Thanks, but can you wait with merging this a little bit? I am working on integration of virtio pipes into the network directory (so they can be dialed like network connections) and straightening out some logic per your prior review?

rminnich:
OK, I'll take the .c file out but I think the rest of the includes should stay as is?

I don't see the argument for putting virtio pipes into /net. What does a dial string look like for them? But I'm willing to be convinced.

rminnich:
remember /net is just a convention. tcp!x!y is just a convention -- there's nothing special about it, it's just a pathname component and is interpreted by the ip stack in a special way by devip -- it's why ! was used in the first place. There's also /net.alt. /net is a path, nothing special. So, again, I'm lost on the need to put something /net that looks like a pipe. You can dial ANY path as long as that path follows certain rules. That's why dial takes strings.

So what's a virtio !! string look like? I'm not seeing it yet.

dmgolubovsky:
I thought about 'virtio!mount_tag'. I don't need any userspace utilities then, 9fs will be able to take care of it. I have to provide a "clone" file which would instantiate a connection, then receive the "connect" command to engage an actual virtqueue. This is my understanding of the protocol behind networked connections. A program wiling to work with virtqueues themselves might dial something like virtio!device!queue_number.

There may be other ways to identify the devices behind virtio; mount tags are just one special case. at the very end there is always artificial naming based on pci enumeration like the current devvq does.

I haven't committed any code which does this, currently coding it. So devvq.c will definitely be different. Other includes should be OK.

Thanks.

elbing:
Marked as WIP until author said the opposite (or claims it's ready).

dmgolubovsky:
Thanks.

rminnich:
based on what your discussion above I don't agree with your understanding of network connections. What is the tag in virtio!mount_tag? It should not be a 9p tag as this is not needed. No other 9p service/client has ever needed it. This does not sound like the right direction to me.

rminnich:
What is the connection operation in a virtio queue? What do you have to do to connect it? The connect in devip corresponds to a connect in tcp, i.e SYN-ACK phase. We've seen no need for such a thing in the virtio layer on akaros (based on the same code you're using, btw). I'm really not able to understand what you're saying here but the ! thing doesn't look right to me.

dmgolubovsky:
I am not sure why 9p mount tag is not needed. How else do I capture the host-presented label of a 9p service? I expect having many 9p connections available from the host, so some labeling is necessary anyway. PCI bus enumeration would not help here.

Per connection operation: my vision is that once clone is opened, it creates a new connection's ctl which later will be forced by writing "connect" message to it, and whatever this message has as an argument will pick the proper virtqueue, which will be further used as a pipe for sending packets back and forth (via per-connection data file). This is from my reading of manpages. I want to be able to use a script like 9ph (via calling srv) to mount virtio pipes providing 9p just like any other networked 9p connection. For other use of virtqueues (e. g. virtio GPU) definitely other ways to connect will be needed, but I am looking into 9p right now.

If this direction is not right which one would be? I found what I proposed initially a bit overcomplicated.

rminnich:
In my opinion it's still too complicated. You are putting two things together that do not need to go together, 9p and the pipe-like nature of virtio. I also think your reading of man pages has led to some incorrect conclusions. virtio is a transport, as are devpipe and devip, and none of them need to have these tag operations. 

Let's start from the start. Can we get these include files in as I've got them here and then start talking about devvq and how it should work? We've been through a lot of this on akaros already and simplicity should be the rule. 

The base level virtio rings should give you simple pipe like transport and nothing more. On top of that you can have net, disk, 9p, or whatever. But you really want to keep devvq.c really simple, and there's a lot you can remove from it even now (all the ownerpid code for example).

The progression I'd like to see is this:
get these include files in
very simple devvq.c that just does correct resource discovery

then we talk about what ops devvq should support, i.e. namespace it presents and what those ops mean. At every step it's small changes with working code and a design we all understand.

dmgolubovsky:
Sure, let's add the include files, I never objected to that. Or even add devvq.c if it does not conflict with anything else; I'll start cleaning it up.

Thanks.

dmgolubovsky:
OK, let's merge as proposed and I'll take it from there.

Thanks.

rminnich:
I left the attributed packed stuff in even thought that's not really a good style. Let it go for now.

LGTM per golubovsky

PR #309, opened 2016-09-20
sys/src/cmd/samterm: fix suicide on mouse movement

hdonnay:
In io.c, put `Mouse *` in Alt.v instead of `Mouse **`

Change-Id: I3eb18146b2b770fc466259b8cdc76f9983049791
Signed-off-by: Hank Donnay <hdonnay@gmail.com>

Also hit the file with `clang-format`. Actual changes occur on lines 134 & 151.

### DIFF
--- a/sys/src/cmd/samterm/io.c
+++ b/sys/src/cmd/samterm/io.c
@@ -17,24 +17,24 @@
 #include "flayer.h"
 #include "samterm.h"
 
-int	cursorfd;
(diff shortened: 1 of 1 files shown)


hdonnay:
Samterm still leaves its terminal hosed, but now it's at least usable until you quit. I can't recall if it works when a file is plumbed to it.

elbing:
LGTM

PR #313, opened 2016-09-27
Add a bind regression test (it fails)

rminnich:


rminnich:
it now works with the fix to ninep

elbing:
LGTM

PR #314, opened 2016-09-29
Remove weird $HARVEY check in build

floren:
build was checking if you have HARVEY set and, if so, *not*
adding util/ to your PATH. Fix that because it caused weird
problems.

### DIFF
--- a/util/src/harvey/cmd/build/paths.go
+++ b/util/src/harvey/cmd/build/paths.go
@@ -11,9 +11,6 @@ import (
 
 func init() {
 	harvey = os.Getenv("HARVEY")
-	if harvey != "" {
-		return
-	}
 	// git is purely optional, for lazy people.
 	out, err := exec.Command("git", "rev-parse", "--show-toplevel").Output()
 	if err == nil {
--- a/util/src/harvey/cmd/build/paths.go
+++ b/util/src/harvey/cmd/build/paths.go
@@ -11,11 +11,14 @@ import (
 
 func init() {
 	harvey = os.Getenv("HARVEY")
-	// git is purely optional, for lazy people.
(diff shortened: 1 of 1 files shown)


util/src/harvey/cmd/build/paths.go
```diff
@@ -11,9 +11,6 @@ import (
 
 func init() {
 	harvey = os.Getenv("HARVEY")
-	if harvey != "" {
-		return
-	}
 	// git is purely optional, for lazy people.
```
hdonnay: exec'ing git should go in a `harvey == ""` conditional

util/src/harvey/cmd/build/paths.go
```diff
@@ -11,9 +11,6 @@ import (
 
 func init() {
 	harvey = os.Getenv("HARVEY")
-	if harvey != "" {
```
rminnich: I'm not so sure about this. I think the intent is that if you are running from a not-git place it will work. I'm actually willing to yank this whole function, it assumes you have git installed. Andy why can't people just set PATH and HARVEY without crutches?

I don't know enough to LGMT.

floren: The intent is this:

First we grab $HARVEY if it exists.
Then we try to get the root of the current tree from git. If that works (without error!) we use the result of git to overwrite what we read from $HARVEY
If we get past the git command and the harvey variable still isn't set, we admonish the user to set $HARVEY and bail out.

This code allows whatever git finds to supersede $HARVEY. We could do it the other way around, too.

I think it's acceptable to require people to set either HARVEY or PATH but not both...

hdonnay:
LGTM

Issue #324, opened 2016-10-14
Harvey only works with 9P2000, QEMU virtio wants 9P2000.u or .L


dmgolubovsky:
I got some progress with virtio, but run into a problem of 9P version incompatibility with QEMU.

The 9P version that Harvey supports is hardcoded in fcall.h

```
#define VERSION9P       "9P2000"
```

In QEMU, (hw/9pfs/virtio-9p.c) the only versions recognized are:

```
    if (!strcmp(version.data, "9P2000.u")) {
        s->proto_version = V9FS_PROTO_2000U;
    } else if (!strcmp(version.data, "9P2000.L")) {
        s->proto_version = V9FS_PROTO_2000L;
    } else {
        v9fs_string_sprintf(&version, "unknown");
    }

```

If I try to mount a virtio connection I get error "bad 9P version returned from the server". If I change the string in fcall.h to 9P2000.u everything else dies.

The mount syscall does not currently accept any flags that could be passed to devmnt to adjust the negotiated version of 9P. What could be the best way to handle this?

Thanks.

rminnich:
I think we need to find whatever person broke QEMU 9p support, since they clearly don't understand how 9p works. There is NO REASON for that requirement, unless you can prove otherwise. Or I need to see a convincing reason that harvey should add 9p2000.[uL] support, which support was ONLY intended for unix systems, which harvey is not.

So the right way to handle this is to fix QEMU. Harvey's not broken. QEMU is.

Also, I think it's a good idea to get your changes up as PRs. Reviews are a good way to make sure you're staying on the right trail. If you're changing VERSION9P in harvey, you are definitely on the wrong trail.

rminnich:
This not a harvey bug. This is a qemu limitation. You can only fix this by writing a new devmnt for 9p2000.l, and the writing all the RPC support too. I'm going to guess that's more than you planned. If you want a working 9p2000 support from a VMM, try https://github.com/penberg/linux-kvm/blob/e325f3e77e782a561e5c53ad582b0ac088573a59/net/9p/client.c#L98

But I'm curious why you're trying 9p on virtio before you get the virtio support in to harvey.

rminnich:
I talked to my QEMU contact and basically they don't support 9p2000 because they did not need it for linux clients. There's no pushback on fixing QEMU. So I'd suggest you focus on virtio-net for now, ignore virtio-9p, and I'll try to fix QEMU.

rminnich:
[...]

PR #180, opened 2016-06-09
Bring in the virtqueue definitions from the virtio spec 1.0. Make sure BAR address and size are picked up for virtio-pci devices with ID=0x00 (e. g. virtio-9p).

dmgolubovsky:
Next small step in the virtio adoption. The virtqueue definitions brought in from the virtio spec 1.0 site. The virtio_ring.h file has been #include'd into devpci.h just to make sure it compiles for now. It has been noticed that the when the PCI devices are scanned, devices with ID=0x00 are skipped, so pci(8) does not display BARs and sizes for them. Devices like virtio-9p belong to this category, so now their information is displayed more accurately.

Before this change (pci(8) output from the previous pull request):

0.0.0:  brg   06.00.00 8086/1237   0
0.1.0:  brg   06.01.00 8086/7000   0
0.1.1:  disk  01.01.80 8086/7010   0 4:0000c2a1 16
0.1.3:  brg   06.80.00 8086/7113   9
**0.10.0: virtio-balloon 00.ff.00 1af4/1002  10**
0.11.0: virtio-disk    01.00.00 1af4/1001  11 0:0000c1c1 64 1:fc098000 4096
0.2.0:  vid   03.00.00 1b36/0100  10 0:f4000000 67108864 1:f8000000 67108864 2:fc090000 8192 3:0000c201 32
0.3.0:  net   02.00.00 10ec/8139  11 0:0000c001 256 1:fc092000 256
0.4.0:  virtio-net     02.00.00 1af4/1000  11 0:0000c221 32 1:fc093000 4096
**0.5.0:  virtio-9p      00.02.00 1af4/1009  10**
0.6.0:  virtio-console 07.80.00 1af4/1003  10 0:0000c241 32 1:fc095000 4096
0.7.0:  virtio-scsi    01.00.00 1af4/1004  11 0:0000c141 64 1:fc096000 4096
0.8.0:  virtio-disk    01.00.00 1af4/1001  11 0:0000c181 64 1:fc097000 4096
**0.9.0:  virtio-rng     00.ff.00 1af4/1005  10**

After the change:

0.0.0:  brg   06.00.00 8086/1237   0
0.1.0:  brg   06.01.00 8086/7000   0
0.1.1:  disk  01.01.80 8086/7010   0 4:0000c2a1 16
0.1.3:  brg   06.80.00 8086/7113   9
**0.10.0: virtio-balloon 00.ff.00 1af4/1002  10 0:0000c281 32**
[...]

PR #325, opened 2016-10-15
Add somewhat working primary virtio queue support.

dmgolubovsky:

This PR adds primary support functions for the virtqueue transport. Common functionality provides the namespace for virtqueues as files that can be opened for sending and receiving data. Each type of a virtual device needs a specific "flavor" functions to be written (for reading, writing, and cleaning up device-specific data at the moment; specialized initialization functionality can be embedded in the flavor's "constructor" function). Flavor for 9p is provided, but per the recent discussion (bug #324) it is unlikely to work properly with the present QEMU implementation of 9p. Also provided are basic functions to borrow virtio buffer descriptors from the queue, placing them in the queue, and handling interrupts. Testing was limited to the attempts to implement 9p.


sys/src/9/amd64/devvq.c
```diff
@@ -0,0 +1,638 @@
...
+// Map device identifiers to descriptive strings to display as virtio device names.
+
+typedef struct
+{
+	uint16_t did;
+	char *desc;
+	Flavor *flv;
+} didmap;
+
+static didmap dmtab[] = {
```
rminnich: The virtio spec defines a basic transport. I think there's a bit too many things in this file. The plan 9 way to implement this type of model would be to define the basic virtio transport, and then define the types of transports that use it as different entities. Consider the way that tcp/ip stack works. There are ethernet devices, and there are network stacks. To get a stack to use a device, you do something like
echo ether0 > /net/ipifc/0/ctl
I think there are too many levels squished in here. I think that is indicated by the fact that you need this dmtab -- it just doesn't feel right. Could you maybe just implement the the basic virtio queue abstraction and then we can look at implementing the various tranports in other drivers? That's more the style of plan 9. 
Or, maybe, look at the storage device abstraction (devsd) and how the different types of drivers (sdata, sdscsi, etc.) link into it. This is really great progress but the presence of this dmtab just doesn't seem right.

or just consider the relationship of #$ (pci device) to PCI devices. #$ is a generic driver for managing the pci abstraction, and individual devices might use it.

sys/src/9/386/pci.c
```diff
@@ -193,7 +193,6 @@ pcilscan(int bno, char *path, Pcidev** list)
 				}
 				break;
 
-			case 0x00:
```
dmgolubovsky: This was part of PR #180 (Some virtio-pci devices (e. g. 9p)  have ccrb = 0x00, their BARs and sizes also shouldbe  picked up here) which was reverted (probably by omission). It has to do with virtio of course.

sys/src/9/amd64/vq9p.c
```diff
@@ -0,0 +1,216 @@
+/*
```
dmgolubovsky: It is probably intel-specific as is all virtio (does Linux on ARM provide identical functionality in terms of virtualization, and is it in the same wide use as on amd64)?

sys/src/9/amd64/devvq.c
```diff
@@ -0,0 +1,638 @@
...
+
+#include	"virtio_custom.h"
+
+#define TYPE(q)			((uint32_t)(q).path & 0x0F)
+#define DEV(q)			((uint32_t)(((q).path >> 4) & 0x0FFF))
+#define VQ(q)			((uint32_t)(((q).path >> 16) & 0x0FFFF))
+#define QID(c, t)		((((c) & 0x0FFF)<<4) | ((t) & 0x0F))
+#define VQQID(q, c, t)	((((q) & 0x0FFFF)<<16) | (((c) & 0x0FFF)<<4) | ((t) & 0x0F))
+
+#define BY2PG PGSZ
```
dmgolubovsky: is this about BY2PG? I think it is one remnant from 9front code which used this constant, can be changed.

sys/src/9/amd64/devvq.c
```diff
@@ -0,0 +1,638 @@
...
+	Qvqdata,				// virtqueue file entry for reading-writing messages
+	QVqctl,					// virtqueue file entry for control operations, write-only
+};
+
+enum
+{
+	Vqqsize = 64*1024		// default size for the device read queue size
+};
+
+static Dirtab topdir[] = {
```
dmgolubovsky: I would not unify them with #| simply because as we see in the 9p example they may be far from dumb pipes. Virtio-net has command and data queues, and they make no sense to be used separately. Same with virtio-sd, they are specialized. Some virtio devices may open their queues for direct read-write, some not. My current code (via per-device flavor) allows for unified device identification and discovery, but if the flavor implementation blocks read/write access to virtqueues as simply files (pipes) then they will not be accessible from regular processes (though visible as device metadata).

dmgolubovsky: Or maybe I misunderstand the question about #|, can you clarify?

sys/src/9/amd64/vq9p.c
rminnich: yeah. We've just finished a port of virtio to akaros and there's almost nothing architecture dependent in it. Is there anything in your code that is x86 specific? I don't see anything.

sys/src/9/amd64/devvq.c
```diff
@@ -0,0 +1,638 @@
+/*
```
rminnich: I don't see anything about virtio that is architecture-specific. Note that rusty russell had a lot to do with virtio and at that time he was an IBM Power guy.

sys/src/9/amd64/devvq.c
rminnich: it needs to be changed.

sys/src/9/amd64/devvq.c
rminnich: you misunderstood the question. 
#| is a simple bidirectional pair of pipes. It's a nice abstraction. 

If we ignore implementation details, virtio is a similar abstraction. I think you ought to be able to define a basic abstraction, which is virtio, that is similar in nature to pipes, in that its function is to provide a way to move data. 

What i'm trying to say is that I'd like to see this PR implement the basic virtio abstraction, i.e. moving data, and then we can add the flavors in a later PR. I don't think that should all be merged in the code and in this PR.

sys/src/9/amd64/devvq.c
rminnich: I think almost all of this code can be kept. If you want I can make some simple changes to implement what I'm talking about.

rminnich: OK, think about it this way. The pipe device provides two bidirectional queues. Why can't the virtio device, similarly, provide a single bidirectional queue? That's a basic abstraction -- a queue. That's how it is intended to work per the spec. You can then build on that abstraction and implement other things, such as net, 9p, console, disk, etc. 

So how about we start with something that implements just the queue?

rminnich: Also, your vqlib idea is pretty clever.

sys/src/9/amd64/vq9p.c
dmgolubovsky: Not x86 specific in the code, but my understanding was that virtio is mainly available and usable on x86 platforms (s390 which is also mentiond in the spec left alone). Plus, it relies on pci, and pci.c is in 386, though devpci.c is in port. That made me think about intel-specificity of virtio-pci. I just cannot see anything like pci.c in aarch64...

sys/src/9/386/pci.c
dmgolubovsky: Perhaps I need to send a separate PR on this?

sys/src/9/amd64/devvq.c
dmgolubovsky: OK, my research in the implementation of virtio (esp. how 9p, sd, and net are implemented - they do not expose any "simple" queues per se) made me think that this is somewhat different from just a pipe (although very similar). I'll try to study the logic of #| and what was done in Akaros. I'll push the code for vqlib (these few functions as much as I can separate them) to place them in port, just need to understand better what handles PCI for ARM in Harvey.

Thanks.

rminnich:
My point here is we don't need to limit our thinking. The virtqueue is a nice construct, as it is designed to work in a very efficient manner. The virtqueue, considered alone, could be useful for many things. We should not limit our idea about possible uses to what the linux/kvm world is using it for. Also, the way things are done in plan 9 is not at all the way they're done in the unix world, so we need to integrate this new construct into harvey in a way that makes sense for harvey. 

I don't think this requires lots of changes in what you have here; most of what I'm advocating is a removal of some of the code, but most of this PR looks really good.

as for PCI, the reasons that some of pci is in 386/ is that the 0xcf8/0xcfc configuration space mechanism is very x86 specific. But devpci can be in port/ because a lot of pci is NOT x86 specific. Similarly, I still feel that all the code you have written can go in port/ instead of amd64/.

dmgolubovsky:
This PR is not going to proceed, so closing it now.

PR #328, opened 2016-10-17
Initial commit of the virtio library.

dmgolubovsky:

This is the abstract virtio code extracted from former devvq. One addition needs to be done elsewhere: the initvdevs function (find and initialize all virtio devices) used to be called from devvq's devinit method. Since we no longer expose all queues via a device, there should be place in Harvey's initialization where such function can be called.

### DIFF
--- a/sys/src/9/port/virtio_lib.c
+++ b/sys/src/9/port/virtio_lib.c
@@ -1,0 +1,283 @@
+/*
+ * This file is part of the Harvey operating system.  It is subject to the
+ * license terms of the GNU GPL v2 in LICENSE.gpl found in the top-level
+ * directory of this distribution and at http://www.gnu.org/licenses/gpl-2.0.txt
+ *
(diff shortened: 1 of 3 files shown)


sys/src/9/port/virtio_lib.c
```diff
@@ -0,0 +1,283 @@
...
+ * This file is part of the Harvey operating system.  It is subject to the
+ * license terms of the GNU GPL v2 in LICENSE.gpl found in the top-level
+ * directory of this distribution and at http://www.gnu.org/licenses/gpl-2.0.txt
+ *
+ * No part of Harvey operating system, including this file, may be copied,
+ * modified, propagated, or distributed except according to the terms
+ * contained in the LICENSE.gpl file.
+ */
+
+// devvq.c ('#Q'): a generic virtqueue driver.
```
rminnich: do we need this any more?

PR #330, opened 2016-10-20
apex: add a program to test how apex exits

rminnich:


rminnich:
not needed, so we can delete this if we don't want it.

elbing:
Not sure... When would be an installation iso or anything like that, toolchain, Go and APEX will be in it by default, with the possibility of being updated through git. But for now, that we must mount the system by pieces... As you wish.
Needs a rebase, anyway.

elbing:
Thought better, if it's not needed, delete it.

rminnich:
not needed -- it was a simple apex bug and this doesn't really test anything interesting.

PR #332, opened 2016-10-23
riscv: this actually prints a panic message.

rminnich:


elbing:
rebase.

rminnich:
but if you COULD review this and see if you can approve it that would help. I want to keep up with HEAD.

PR #333, opened 2016-10-25
riscv: booting to a bsp boot function

rminnich:
A lot more is working. We can link with klibc and use memset.

We can set up initial mach pointer and stack in C, not assembly, which
avoids lots of nastiness.

Next step: split amd64/asm.c in arch and port parts, so we can use that code.

### DIFF
--- a/sys/src/9/riscv/main.c
+++ b/sys/src/9/riscv/main.c
@@ -18,13 +18,61 @@
 
 void testPrint(uint8_t c);
 
-void fuck(char *s)
+void msg(char *s)
 {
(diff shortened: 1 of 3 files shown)


sys/src/9/riscv/main.c
```diff
@@ -18,23 +18,70 @@
 
 void testPrint(uint8_t c);
 
-void fuck(char *s)
+void msg(char *s)
```
elbing: ha ha! I liked original name of the function :-p

PR #334, opened 2016-10-25
riscv: get all the libs and cmd to build

rminnich:
We have to stop building snoopy and the aeexec tool by
default for now (you can still build them for amd64 by hand).

But with this change you can build at top level and everything
builds as expected. For instance, building riscv for me is this:

ARCH=riscv CC=~/projects/riscv/riscv-tools/bin/riscv64-unknown-elf-gcc\
	LD=~/projects/riscv/riscv-tools/bin/riscv64-unknown-elf-ld build

and it builds. The kernel no longer links but that's taken care of by sweat equity.

Harvey is once again multi-architecture, at least for libraries and userland.


sys/src/libmach/6.c
```diff
@@ -10,10 +10,10 @@
 /*
  * amd64 definition
  */
-#include <u.h>
+#include <amd64/include/u.h>
```
elbing: it would be better if libmach json could have a cflag -I/$ARCH/include. Not only for eye candy, better for portability. No?

rminnich: I think you're missing the point. libmach is compiled for all architectures. So if you're going to compile 6.c for the riscv, then you need to explicitly include amd64/include/u.h in 6c to get the right register set. The -I /$ARCH is already set up and working, and if you just use u.h for 6.c, you get the riscv register set definitions instead of the ones you want for 6.c. 

For each architecture type in libmach, e.g. 6.c, you have to specify the full path of the u.h, since the u.h might be for a different ARCH than the one you're on. Hence this required change.

rminnich:
anyone?

PR #335, opened 2016-10-25
writev: simple implementation [WIP]

rminnich:
I hope this is good enough for Apex to start.


elbing:
I'll try this code in APEX lib and if it works, we could merge this PR. It's important, because writev now it's the main __stdio_write method for writing standard output inherited from musl.

rminnich:
this is not really needed either.

PR #338, opened 2016-10-26
Compile minimal aarch64 kernel.

dancrossnyc:
Where 'kernel' is defined as a single 'WFI' instruction.


elbing:
Doesn't matter, Unix just said hello world the first time Ken booted it...

dancrossnyc:
Exactly; minimal instruction followed by some prints and then get down to the meat of things.

PR #341, opened 2016-10-27
riscv: get it to build

rminnich:
It won't remotely work, but I've found it easiest to at least get something to build
and then take it from there.


elbing:
_tas32?

rminnich:
thanks.

PR #344, opened 2016-10-28
mv and cp for compatibility and life easy

elbing:
Change-Id: Ia4f47a25b48d49661eee2f7b0d888eb0540305ee

### DIFF
--- a/sys/src/cmd/mv.c
+++ b/sys/src/cmd/mv.c
@@ -53,6 +53,14 @@ main(int argc, char *argv[])
 		exits("bad usage");
 	}
 
+	/* Skip -f */
+	if(argv[1][0] == '-' && argv[1][1] == 'f' && argv[1][2] == 0) {
+		for(i=2; i<argc; i++) {
+			argv[i-1] = argv[i];
+		}
+		argc--;
+	}
+
 	failed = 0;
 	for(i=1; i < argc-1; i++)
 		if(mv(argv[i], todir, toelem) < 0)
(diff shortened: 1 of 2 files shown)


elbing:
I could add a mv for apex, but... well this is the simplest case. This allows not changing hundreds of lines in hundreds of GNU configure scripting.

sys/src/cmd/mv.c
```diff
@@ -53,6 +53,14 @@ main(int argc, char *argv[])
 		exits("bad usage");
 	}
 
+	/* Skip -f */
+	if(argv[1][0] == '-' && argv[1][1] == 'f' && argv[1][2] == 0) {
```
dancrossnyc: just use ARGBEGIN/ARGEND here? it's probably less code....

elbing: It wasn't in original code because mv hasn't any flag on Plan 9. This ensures the first argument is nor passed as argument, so the classic and cool way... maybe is not appropiate. This is an ugly patch, officially mv hasn't flags. The way you like would be like:

    ARGBEGIN{
    case ’f’:
        break;
    default:
        usage();
    }ARGEND

Do you really want so? no problem for me, cp has it in that way because there were previous flags and I just added "p".

dancrossnyc: I do think I like it better. Another option would be to have some compat/bin directory that gets bind -b'ed onto /bin, that could just remove the '-f' option and invoke the real /$cputype/bin/mv:

#!/bin/rc
if ( ~ $1 '-f' ) { shift }
exec /$cputype/bin/mv $*

elbing:
This will be a couple of shell scripts emulating this flags for not polluting harvey's code with dead code.

PR #345, opened 2016-10-28
riscv: remove getcaller[pc,stack].S

rminnich:
Dan gave us portable versions.


dancrossnyc:
Does the build file need to be edited as well?

PR #346, opened 2016-10-28
getcallerpc/getcallstack: move common code to libc

rminnich:
Dan brought in nice standard code but pointed out it was
being replicated. This change allows us to use the common
code for cases it can be used. For special cases,
one can still create special versions in the libc/$ARCH
directory. Which one is used depends on libc/$ARCH/build.json.


sys/src/libc/riscv/build.json
```diff
@@ -4,8 +4,8 @@
 	    "$ARCH/notejmp.c",
 	    "$ARCH/argv0.c",
 	    "$ARCH/atomic.S",
-	    "$ARCH/getcallerpc.c",
-	    "$ARCH/getcallstack.c",
+	    "getcallerpc.c",
```
dancrossnyc: These files should go to libc/port, right?

sys/src/libc/amd64/getcallerpc.c
```diff
@@ -1,17 +0,0 @@
-/*
```
rminnich: moved. hmm.

sys/src/libc/riscv/build.json
rminnich: good idea.

PR #347, opened 2016-10-28
Rewrite cistrcmp and strcmp.

dancrossnyc:
Make strcmp and cistrcmp more straight-forward.  I believe
the originals were correct, but these are more obvious.
Note: Please review these *very* carefully.

### DIFF
--- a/sys/src/libc/port/cistrcmp.c
+++ b/sys/src/libc/port/cistrcmp.c
@@ -9,27 +9,25 @@
 
 #include <u.h>
 #include <libc.h>
+#include <ctype.h>
 
 int
 cistrcmp(const char *s1, const char *s2)
 {
-	int c1, c2;
+	unsigned int c1, c2;
 
-	while(*s1){
+	do{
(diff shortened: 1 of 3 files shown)


sys/src/libc/port/cistrcmp.c
```diff
@@ -9,27 +9,27 @@
...
+	unsigned int c1, c2;
 
-	while(*s1){
+	do{
 		c1 = *(uint8_t*)s1++;
 		c2 = *(uint8_t*)s2++;
-
 		if(c1 == c2)
 			continue;
+		if(isupper(c1))
```
rminnich: were I a betting man, I'd bet there is no difference in cost to just do the tolower here and skip the test. bye bye branch, and so on.

dancrossnyc: You are correct: I just checked toupper.c (which also has tolower.c in it) and it does a check itself so we should be good to go.

elbing:
Clever!

PR #348, opened 2016-10-28
Move devether.c to port.

dancrossnyc:
devether is the glue that ties the IP stack
to ethernet devices; it's essentially a kernel
API with a filesystem interface.

There is some suspect, possibly 386-only code
in the probe routine having to do with IRQ
selection, but if it turns out to be a problem
on other architectures, we can fix that.

Signed-off-by: Dan Cross cross@gajendra.net


rminnich:
yeah but it's more than just ip stack, which is why it's in 386/. I think putting it in ip/ is not quite right.
maybe it could go in port/, and break out the architecture dependent bits into arch directories, somewhat like the port/devpci.c and 386/pci.c split.

elbing:
Yes, the dev/* things are in port. In fact it should be even independent of ip stack if we change it in future as we thought a year ago.
I think things in 386 are not really 386, are all amd64. Unless we consider that dir as generic_x86. But the name moves to a misunderstanding. Over all in a system that doesn't support 32 bit x86.

dancrossnyc:
I like putting it into port makes sense, and then having an arch specific 'ether.c' in each directory.

Regarding Alvaro's point, perhaps we should just get rid of 386; we're not going back to supporting 32-bit x86, so having that directory as "these are the things shared between x86 and amd64" doesn't make a lot of sense.

Similarly, I'm kind of in favor of moving ip/* to port. I've never seen much of a point in keeping the IP stack separate. But that's another change....

elbing:
I think ip keeps its dir since it went back to kernel. It has the advantage of "hey, this is all ip stack related for kernel", but yes, it could be a subdir of port. Maybe sys/src/9 should be kernel, krn or so. Libc/9sys and 9syscalls loose its 9 in the name, which is identified with 9P this days or under the curse. But those cosmetic changes can be discussed when we could make an iso. And for that, harvey must buiild itself and all what's needed to make an iso with a full running system. We're close, but not today.
Anyway, moving 386 to amd64 and ip under ports (not merging their contents), ok for me. Maybe one day we should make dirs for drivers, vga and so on.

dancrossnyc:
We're a bit far afield from devether now, but I kind of like keeping '9' around in the names. To me, it feels like homage, and I like the aesthetics of it.

Aesthetic appeal is one of the things that I find compelling about Plan 9, and by extension Harvey.

Of course, there is also the philosophy of the system as a whole. There is a sort of reformation and distillation of the essence of Unix in Plan 9 (and derived systems) that's been lost in Linux and modern Unix-like systems (e.g., the BSDs, etc).

elbing:
Ok, I buy it. 9 homage, I mean. /port/ip then? We could wait s little bit niw the new build tool is coming and we should begin replacement operation. It will be the fine chance for changing the kernel tree as we want.

dancrossnyc:
port/ip maybe, or just ip, or even combining the contents of ip into port. I'm not sure about that; I figure: one step at a time. For the time being, I think devether.c belongs in port. Let's move this file and we can debate the rest. :-)

sys/src/9/port/riscvport.json
```diff
@@ -27,6 +27,7 @@
 			"../port/devcons.c",
 			"../port/devdup.c",
 			"../port/devenv.c",
+			"../port/devether.c",
```
rminnich: does this work? I'm ok with not putting it here until I test it.

dancrossnyc:
Sure. It's just moving a file from directory 'A' to directory 'B'; no other changes. As long as it builds and links it's fine. (Though in fairness, those may be famous last words.)

rminnich:
yeah but we don't even attempt 386/ in riscv/

elbing:
rebase. For me ok.

dancrossnyc:
Ron is right; we need to split this up.

I'm going to go ahead and close this PR and work on cleaning up the driver while it's in 386, then we can move it to port.

PR #349, opened 2016-10-29
riscv: get further in the boot process [WIP]

rminnich:
we have print. Now need to work out meminit.

Signed-off-by: Ronald G. Minnich rminnich@gmail.com


sys/src/9/riscv/main.c
```diff
@@ -119,14 +120,34 @@ void bsp(void)
 
 	static Lock l; // to ensure initialization.
 	if (canlock(&l)) msg ("L can be locked\n"); else die("Can't lock L\n");
-	ilock(&l);
+	unlock(&l);
 	if (canlock(&l)) msg ("L can be NOT be locked OK\n"); else die("Can lock L after lock\n");
```
dancrossnyc: "L can be NOT be locked OK\n" <-- I think that's got an extra 'be': how about, "L can NOT be locked OK\n"

dancrossnyc: Or even, "L can NOT be locked after lock\n"

dancrossnyc: (Or, "L can NOT be locked while held\n")

rminnich:
you can kind of ignore this for now, I am using github as a storage device.

elbing:
wow, this looks epic.

PR #350, opened 2016-10-30
Fix BUILD files for the new build tool

rminnich:
Also add a bit of info to bootstrap.sh.

The new build tool is still not useable but at least people can see
how nice it's going to be.


sevki:
isn't this the same PR I have? :)

PR #353, opened 2016-10-31
Clean up and split devether.c into portable and arch-specific parts.

dancrossnyc:


sys/src/9/port/devether.c
```diff
@@ -20,11 +20,14 @@
...
 
+extern Ether *archetherprobe(int ctlrno, char *type, int (*reset)(Ether *));
+extern void archethershutdown(Ether *ether);
+
 Chan*
 etherattach(char* spec)
 {
 	Proc *up = externup();
-	uint32_t ctlrno;
+	int ctlrno;
```
rminnich: but shouldn't this be unsigned somehow? it's an array index.

sys/src/9/port/port.json
```diff
@@ -21,6 +21,7 @@
 			"../port/devcoreboot.c",
 			"../port/devdup.c",
 			"../port/devenv.c",
+			"../port/devether.c",
```
rminnich: if you put this in, e.g., amd64/whatever.json, we would not need these fake ethernet devices in riscv/. That seems better to me.

sys/src/9/riscv/ether.c
```diff
@@ -0,0 +1,35 @@
+/*
```
rminnich: I'd be happier not to have this. Another option is to put an
ethernet.json
in port/ and only have amd64 use it.

sys/src/9/port/devether.c
dancrossnyc: I don't think so; not necessarily anyway. There's a long precedent for using variables of type `int` as a loop counter to index into an array.

dancrossnyc: Btw: if it were global or meant to be persistent I'd generally agree; but it's not.

sys/src/9/port/port.json
dancrossnyc: Ok. I think I've done that.

sys/src/9/riscv/ether.c
dancrossnyc: I put this in amd64/core.json; there are a few devs for which this is done. Just to be clear, though, you want this out of riscv for now just to avoid interface churn? That's totally fine; I'm just trying to get ahead of what'll come down the pike when you're ready to bring up an ethernet interface.

PR #354, opened 2016-11-01
Virtual serial console somewhat working.

dmgolubovsky:
Somewhat working virtual console. To test (neither works completely, but shows that communication is working), create 2 virtual serial consoles:

```
-device virtio-serial-pci,max_ports=1 \
-device virtconsole,chardev=vc02 \
-chardev socket,id=vc02,path=/tmp/vc02 \
-device virtio-serial-pci,max_ports=1 \
-device virtconsole,chardev=vc00 \
-chardev socket,id=vc00,path=/tmp/vc00,server,nowait \

```

Note that virtio-serial-pci is specified twice. I decided not to complicate the logic with the multiport feature support, but this means that for several consoles we need several buses (which is provided by virtio-serial-pci). Only port 0 is supported, which means that only "virtconsole" can be used (not sure if there is any difference between virtconsole and virtserialport).

Run ufs on one of them (before starting QEMU so ufs listens already):

```
$HARVEY/util/ufs -ntype=unix -addr=/tmp/vc02 -root=/tmp
```

Connect terminal input to another:

```
socat stdio unix-connect:/tmp/vc00
```

The virtual consoles will be visible as:

```
--rw-rw-rw- C 0 harvey harvey 0 Feb  7  2006 /dev/virtcon/vcons0
--rw-rw-rw- C 0 harvey harvey 0 Feb  7  2006 /dev/virtcon/vcons1
```
Try to mount vcons0 via 9p. The resulting message will be:

```
bad fversion conversion on reply
```

This looks the same as in the prior experiments with virtio-pci: mount asks for a buffer size 8192 bytes, and gets this number of bytes as the result of read, but expects a much shorter message. We can discuss this later, this just shows that data go back and forth over the virtqueue.

Try to communicate with the attached terminal on the host:

```
echo 123 >/dev/virtcon/vcons1
```

will show 123 in the terminal running socat. The reverse however does not work very well:

```
cat </dev/virtcon/vcons1
```

again, it requests an input buffer large enough. If 123<enter> is typed in the host terminal, the whole buffer is returned, showing 123 in the beginning, and the rest is garbage.

So, we have raw communication working primarily. In order to make some protocols work over it, we may need additional filter program which for the case of 9p will try to determine the length of the message received by decoding its header, and pass only the necessary number of bytes. For terminal input, assuming it is in line-mode, such filter may look for the line end symbols and discard the rest.

This is of course not sufficient level of testing; I suggest that after this is merged, more people try to play with this.

Thanks.

### DIFF
--- a/sys/src/9/port/virtio_lib.c
+++ b/sys/src/9/port/virtio_lib.c
@@ -26,7 +26,56 @@
 
 #include	"virtio_lib.h"
 
-#define MAXVQS 8 			// maximal number of VQs per device
+#define MAXVQS 8 			// maximal detectable number of VQs per device
+
+static uint32_t nvq;		// number of the detected virtio9p devices
+
+static Vqctl **cvq;			// array of device control structure pointers, length = nvq
+
+// Map device identifiers to descriptive strings to display in IO port
+// and interrupt allocation maps.
+
+typedef struct
+{
+	uint16_t did;
+	char *desc;
+} didmap;
+
+static didmap dmtab[] = {
+	PCI_DEVICE_ID_VIRTIO_NET, "virtio-net",
+	PCI_DEVICE_ID_VIRTIO_BLOCK, "virtio-block",
+	PCI_DEVICE_ID_VIRTIO_BALLOON, "virtio-balloon",
+	PCI_DEVICE_ID_VIRTIO_CONSOLE, "virtio-console",
+	PCI_DEVICE_ID_VIRTIO_SCSI, "virtio-scsi",
+	PCI_DEVICE_ID_VIRTIO_RNG, "virtio-rng",
+	PCI_DEVICE_ID_VIRTIO_9P, "virtio-9p"
+};
+
+// Find a device type by its PCI device identifier, used to assign device name in the filesystem,
+// and to determine the flavor of read-write operations.
+
+static didmap *
+finddev(Vqctl *vc)
+{
+	for(int i = 0; i < nelem(dmtab) ; i++) {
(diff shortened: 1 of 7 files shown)


sys/src/9/port/virtio_lib.c
```diff
@@ -26,7 +26,56 @@
...
+// Map device identifiers to descriptive strings to display in IO port
+// and interrupt allocation maps.
+
+typedef struct
+{
+	uint16_t did;
+	char *desc;
+} didmap;
+
+static didmap dmtab[] = {
```
dmgolubovsky: These numbers are in compact range, but do not start at 0 (0x1000 to 0x1009), not consecutive (RNG is 0x1005, 9P is 0x1009), and can be extended (e. g. there is virtio GPU which has its number, but it may be outside of this range). So instead of a loop (which runs once at device init) we need to add some range check etc. The present version with loop at least seems more open for future extensions.
Thanks.
PS CLANG failed me on a nested function, need to add a patch anyway.

dancrossnyc:
Sorry I didn't get to review this quickly; FYI it seems to break harvey. I'm getting write faults trying to initialize the virtio console. If I disable the initialization routine, I can boot:

```
CPU Freq. 2266MHz
schedinit...
Scan fffffffff0000000
Scan fffffffff00f0000
get_coreboot_info: ret -1
virtio-serial-pci initializing
write fault fail, no segment, pid 1 addr 0xfffffffff4000000 pc 0xfffffffff020af24
panic: cpu0: fault: 0xfffffffff4000000
```

PR #355, opened 2016-11-02
Clean up devether.c

dancrossnyc:
Clean up the logic to make it easier to follow.

### DIFF
--- a/sys/src/9/port/devether.c
+++ b/sys/src/9/port/devether.c
@@ -49,6 +49,7 @@ etherattach(char* spec)
 	if(etherxx[ctlrno]->attach)
 		etherxx[ctlrno]->attach(etherxx[ctlrno]);
 	poperror();
+
 	return chan;
 }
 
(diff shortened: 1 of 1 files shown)


dancrossnyc:
FYI: Tested by booting and running Harvey.

sys/src/9/port/devether.c
```diff
@@ -176,36 +185,40 @@ etheriq(Ether* ether, Block* bp, int fromwire)
...
-				}
-				else
-					ether->Netif.soverflows++;
-			}
-			else
+			if(f->headersonly){
 				etherrtrace(f, pkt, len);
+				continue;
+			}
+			if(fromwire && fx == 0){
```
rminnich: nice. I betcha this was done this way due to a compiler bug somewhere :-)

sys/src/9/port/devether.c
```diff
@@ -245,12 +258,13 @@ etheroq(Ether* ether, Block* bp)
 		splx(s);
 	}
 
-	if(!loopback){
-		qbwrite(ether->oq, bp);
-		if(ether->transmit != nil)
-			ether->transmit(ether);
-	} else
+	if(loopback){
```
rminnich: geez. Wonder why they did stuff like this? Ah right ... loopback was added much later.

sys/src/9/port/devether.c
dancrossnyc: I expect so. Barret and I talked about this once vis-a-vis Akaros, and the idea was that they may have been trying to force more efficient code generation. I think we can get around that by doing an FDO-style build.

PR #357, opened 2016-11-02
Disable panicy functions in devvcon.

dancrossnyc:
Disable calls to functions that panic Harvey at boot.


rminnich:
can you try booting with no virtios in the qemu command line?

PR #363, opened 2016-11-07
Bldy ci

sevki:
add bldy back to the build matrix and into our hearts.


raylai:
Does it still work with #!/bin/sh? I don't see any obvious bash-only syntax, and that would increase portability.

sevki:
@raylai nice catch thanks.

sevki:
@raylai turns out line 15 does. It really doesn't matter since that ci script isn't used anywhere but travis so portability really doesn't matter.

PR #365, opened 2016-11-10
Line and 9p mode of virtconsole transfer

dmgolubovsky:
I added a feature to the virtconsole to process the buffers received from the host so the actual length of data can be returned to the requesting process, rather than just the length of the whole buffer. In line mode (prefill + delimiter) the read buffer is prefilled with an octet that is not expected to be found in the input. By finding the first octet with that value, the driver can determine how much data was actually transferred by the host, and return the value accordingly from read. By prefillling the buffer with 0s and using 0 as delimiter, it becomes possible with certain limitations to run a shell over virtual console. Other mode supported is 9p where length of the message is extracted form the first 4 octets, and it is returned from read.

The following script works:

```
#! /bin/rc -x
echo virtio-console-0 mode 9p > /dev/virtcon/ctl
echo virtio-console-0 maxbuf 131096 > /dev/virtcon/ctl
mount -c -n /dev/virtcon/virtio-console-0 /mnt/xxx

ls -l /mnt/xxx/vc*

time dd -if /dev/random -of /mnt/xxx/txfer1 -bs 1024 -count 1024
time dd -if /dev/random -of /mnt/xxx/txfer1 -bs 1024 -count 10240
time dd -if /dev/random -of /mnt/xxx/txfer1 -bs 10240 -count 10240

ls -l /mnt/xxx/tx*

unmount /mnt/xxx

```

and its output:

```
. /rc/lib/rcmain /tmp/mtst
flag p
finit
flag i
. /tmp/mtst
echo virtio-console-0 mode 9p
echo virtio-console-0 maxbuf 131096
mount -c -n /dev/virtcon/virtio-console-0 /mnt/xxx
ls -l /mnt/xxx/vc00 /mnt/xxx/vc02
--rwxrwxr-x M 28 harvey harvey 0 Nov  9 23:49 /mnt/xxx/vc00
--rwxrwxr-x M 28 harvey harvey 0 Nov  9 23:49 /mnt/xxx/vc02
time dd -if /dev/random -of /mnt/xxx/txfer1 -bs 1024 -count 1024
1024+0 records in
1024+0 records out
0.01u 0.12s 0.50r        dd -if /dev/random -of /mnt/xxx/txfer1 ...
time dd -if /dev/random -of /mnt/xxx/txfer1 -bs 1024 -count 10240
10240+0 records in
10240+0 records out
0.01u 1.60s 4.23r        dd -if /dev/random -of /mnt/xxx/txfer1 ...
time dd -if /dev/random -of /mnt/xxx/txfer1 -bs 10240 -count 10240
10240+0 records in
10240+0 records out
0.01u 3.26s 7.77r        dd -if /dev/random -of /mnt/xxx/txfer1 ...
ls -l /mnt/xxx/txfer1
--rw-rw-r-- M 28 harvey harvey 104857600 Nov  9 23:50 /mnt/xxx/txfer1
unmount /mnt/xxx
exit ''

```

The whole thing is not very stable: repeated use of this script causes kernel fault. I could not test dd for reading from the mounted directory, it crashes the kernel as well.

Anyway, I see this as some progress. It might be interesting to use virtio console as the root filesystem when booting.

Everybody is welcome to test this script and compare their times with mine.

PS I order to access a host directory, ```ufs -ntype=unix -addr=/tmp/vc02 -root=/tmp``` was run; the virtconsole in client mode is redirected to the socket at /tmp/vc02

Thanks.

### DIFF
--- a/sys/src/9/port/devvcon.c
+++ b/sys/src/9/port/devvcon.c
@@ -139,13 +139,22 @@ ctl(void *va, int32_t n)
 
 // Read-write common code
 
+static int32_t
+getmaxbuf(VconData *vc)
+{
+	if(vc == 0)
+		return 32768;
+	else
+		return vc->maxbuf;
+}
+
 static int
 rwcommon(Vqctl *d, void *va, int32_t n, int qidx)
 {
 	uint16_t descr[1];
 	Virtq *vq = d->vqs[qidx];
 	VconData *vc = vq->pqdata;
-	if(n > ((vc!=nil)?vc->maxbuf:32768)) {
+	if(n > getmaxbuf(vc)) {
 		error(Etoobig);
 		return -1;
 	}
@@ -155,7 +164,11 @@ rwcommon(Vqctl *d, void *va, int32_t n, int qidx)
 		error("virtcon: queue low");
 		return -1;
 	}
-	uint8_t buf[n];
+	uint8_t *buf = malloc(n);
+	if(buf == nil) {
+		error("devvcon: no memory to allocate the exchange buffer");
+		return -1;
+	}
 	if(qidx) {
 		memmove(buf, va, n);
 	} else {
@@ -168,8 +181,8 @@ rwcommon(Vqctl *d, void *va, int32_t n, int qidx)
 		q2descr(vq, descr[0])->flags = VRING_DESC_F_WRITE;
 	}
 	int rc = queuedescr(vq, 1, descr);
+	int32_t nlen = n;
 	if(!qidx) {
-		int32_t nlen;
 		if(vc != nil)
 		{
 			switch(vc->vcmode) {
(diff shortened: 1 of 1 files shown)


rminnich:
oh, and ... we have a virtio-9p. I don't see the sense in running 9p over a *CONSOLE*. This seems like a giant addition of technical debt which we really don't want. Sorry! I really appreciate the work you are doing ...

dmgolubovsky:
Ron

Thanks for the review. 

RE line mode. I understand that this is a half-measure, and was done mainly for testing, but since the host never informs the guest how much of the buffer was filled, there has to be some "analysis" done over the received data. It still can be used for some kind of a rescue console though if graphics is not available or failed/stuck.

RE 9p over the console stream. We discussed the problems related to the QEMU implementation of virtio-9p which do not play well with Harvey (protocol versions supported, forced IO operations ordering). What I did at least works with ufs on the other end, and per my measurement gives about 30MB/sec, the same dd (100M random data) over the root mount (via network) took 20 sec or so. And the only difference from virtio-9p is that with virtio console we are not forced to order IO buffers even though this leads to larger number of virtqueue kicks.

Virtio console (or virtio serial port in general) seems to be the most general use pipe between the guest and the host. the rest of the devices (even 9p) add some logic on top of it.

What else use of virtio serial port do you see? It is not byte-stream as provided by QEMU, it is block stream. Byte stream needs some streaming protocol on top of it, it may be either 9p or some kind of point-to-point networking.

Thanks.

rminnich:
But it's called virtio console for a reason. I still feel this is the wrong direction..

dmgolubovsky:
The general framework of which virtio-console is a part is indeed virtio-serial port. But I am not going to argue. I will look how to integrate existing virtio-9p with devmnt, but this will not be as easy as mounting a file handle.

Thanks.

elbing:
Is this replaced for #391?

dmgolubovsky:
Yes, I'm closing this one.

elbing:
Fine! good work, thanks.

PR #368, opened 2016-11-15
Qemu scripts XXI century

elbing:
q35 = ICH9 (not FX440 with pc)
smp 1 is by default, today all the 64 bit processors
have at least two cores.

### DIFF
--- a/util/GO9PTERM
+++ b/util/GO9PTERM
@@ -6,7 +6,7 @@ trap : 2
 $HARVEY/util/ufs -root=$HARVEY &
 ufspid=$!
 
-export machineflag=pc
+export machineflag=q35
 if [ "$(uname)" = "Linux" ] && [ -e /dev/kvm ]; then
 	export kvmflag='-enable-kvm'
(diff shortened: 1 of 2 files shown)


util/GO9PTERM
```diff
@@ -17,7 +17,7 @@ if [ "$(uname)" = "Linux" ] && [ -e /dev/kvm ]; then
 fi
 
 read -r cmd <<EOF
-$kvmdo qemu-system-x86_64 -s -cpu Opteron_G1 -smp 1 -m 2048 $kvmflag \
+$kvmdo qemu-system-x86_64 -s -cpu Opteron_G1 -smp 2 -m 2048 $kvmflag \
```
rminnich: Opteron_G1? That old?

elbing:
What do you want? G5? Phenom? Newer amd cpu in qemu have at least 6 years. Intel nephalem is a 2011 i7 version. Suggestions?

elbing:
The other would be allowing kvm to use native cpu specifying smp. But without kvm you're sold.

PR #370, opened 2016-11-29
bldy: envinroment variables can now be used in paths

sevki:
related to bldy/build#45

Signed-off-by: Sevki <s@sevki.org>

### DIFF
--- a/.build
+++ b/.build
@@ -1,4 +1,5 @@
 CC = gcc
+ARCH=amd64
 [darwin]
 TOOLPREFIX = x86_64-elf-
 [harvey]
--- a/sys/src/libthread/BUILD
+++ b/sys/src/libthread/BUILD
@@ -8,7 +8,7 @@ cc_library(
             "//amd64/include",
         ],
 	srcs = [
-		"amd64.c",
+		"$ARCH.c",
 		"channel.c",
 		"chanprint.c",
 		"create.c",
--- a/sys/src/9/amd64/BUILD
+++ b/sys/src/9/amd64/BUILD
@@ -191,21 +191,21 @@ strip(
 mk_sys(
 	name="error",
 	mode="error.h",
-	arch="amd64",
+	arch=env("ARCH"),
 	sysconf="//sys/src/sysconf.json",
 )
 
 mk_sys(
 	name="sys",
 	mode="sys.h",
-	arch="amd64",
+	arch=env("ARCH"),
 	sysconf="//sys/src/sysconf.json",
 )
 
 mk_sys(
 	name="systab",
 	mode="systab.c",
-	arch="amd64",
+	arch=env("ARCH"),
 	sysconf="//sys/src/sysconf.json",
 )
 
--- a/sys/src/libc/BUILD
+++ b/sys/src/libc/BUILD
@@ -230,17 +230,17 @@ LIBC_SRCS = [
     "port/u16.c",
     "port/u32.c",
     "port/u64.c",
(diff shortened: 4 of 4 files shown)


sys/src/libthread/BUILD
```diff
@@ -8,7 +8,7 @@ cc_library(
             "//amd64/include",
         ],
 	srcs = [
-		"amd64.c",
+		"$ARCH.c",
```
rminnich: Do you need the ${ARCH} form here? Just wondering.

.build
```diff
@@ -1,4 +1,5 @@
 CC = gcc
```
rminnich: I'm not a huge fan of persistent .files. Is this necessary or could we just make it a real name?

sevki: it's not at all necessary but I really don't want to be typing `ARCH=amd64 bldy //:kernel` everytime. contents of this file is overwritten if you specify a real env var like so `ARCH=risc-v bldy //:kernel` bldy will overwrite the ARCH set in .file

sys/src/libthread/BUILD
sevki: I was hoping you'd ask about this, if it's a `.` or `/` terminated, it doesn't seem like you have to. but I can change this to that form for persistence if you'd like.

.build
rminnich: why a .file however? That stuff is a cancer. Why not a file named ENV or DEFAULTS or something reasonably official?

overall however I am not convinced this is a good idea. It's not that hard to
export ARCH=amd64
and then not worry. But it's your tool, up to you, that's why I approved it.
Also, your .build hardwires amd64 and in general hardwire ARCH in any file is bad policy, especially one checked in to git.

sevki: it's not my tool, that's why I created a new org (which is eagerly awaiting you to join) and invited you all.
the .build file comes from the buck and bazel family, I really didn't think too much over it. I'll create an issue and change the .build to something like `bldyconfig`

rminnich: Thank! I really appreciate your taking a look at this!

You can go ahead and commit this PR if you want, the other change can come later, OR you can make the change now. I leave it in your hands :-)

sevki: Ok, I'll merge this now. I have more related to this that will come in but I'd rather avoid making this a cathedral

sys/src/libthread/BUILD
rminnich: I'd suggest making it always work one way. We can see that went really well with Go, people always know there's one and only one way to do things.

sevki: should we check to see if it matches `${.*}` and complain if it doesn't?

rminnich: That regexp is not safe but if you're saying that a $ should be followed by a single {, a set of valid non-} terminated by a }, then you're ok I guess.
${.*} matches ${{{{{{{{{{{{{{{{{{{{{{{{{{{} for example. :-)

sevki: well my question was more about whether we should keep the go functionality as defined [here](https://golang.org/src/os/env.go?s=378:435#L3) or be more opinionated. I'll change it to`${[A-Z]*}`

rminnich: Sorry! I misunderstood.

sevki: no bother, what's your take on being more opinionated?

rminnich: ??

People need to be opinionated :-)

PR #372, opened 2016-12-04
More $ARCH related experimentation

sevki:


sevki:
thanks for pointing it out, I looked at how bazel handles this, and it looks like it uses a simpler notation for sprintf, which I'm guessing comes from python
```
cc_binary_name = "%s.cc_binary" % name
``` 
should we just use that to increase compatibility?

rminnich:
funny you should ask that, I was going to ask you if you'd considered it. Yeah, I would do the python-like thing.

PR #375, opened 2016-12-16
Acpica interrupt support

sanderssj:
This pull request adds support for letting the acpica library, in user space, handle interrupts. This is accomplished by making a real acpi interrupt handler in the kernel that caches the acpica interrupt status register so that when the acpica library reads that address, it gets the cached value. The kernel interrupt handler posts a note a user process that has registered to receive interrupts. Writes from the acpica library affect the cached value (because the kernel interrupt handle has already reset the real status register). This code assumes that every bit in the acpi status register (pm1a)  behaves the same way; i.e. a bit is set when the interrupt has occurred, and gets cleared by writing a 1 to that bit.

The acpica os code is modified to include a note handler to identify the acpi interrupt note and call the acpica handler.

Below you can find an example program which sets handlers for the power button and acpi pm timer interrupts to demonstrate how it works. You can simulate a working power off event with qemu by running the qemu console and using the command 'system_powerdown'. 

I'm sure this code isn't perfect--I think there are ways to miss interrupts and there will need to be synchronization around the pm1a status register. I haven't tried this on real hardware yet either. I have left comments in other areas I'm not sure about.

Hopefully this is at least a starting point for a more robust solution.

testsleep.c:
```
#include <acpi.h>
#include <accommon.h>
#include <acnamesp.h>


void
fail(char *msg)
{
	fprint(2, "%s\n", msg);
	exits("FAIL");
}

UINT32
shutdownhandler(void *ctxt)
{
	ACPI_STATUS status;

	syslog(1, "acpi", "acpi event: power button");
	status = AcpiEnterSleepStatePrep(5);
	if (ACPI_FAILURE(status))
		fail("failed to prep for sleep");
	status = AcpiEnterSleepState(5);
	if (ACPI_FAILURE(status))
		fail("failed to go to sleep");
	return 0;
}

UINT32
timerhandler(void *ctxt)
{
	syslog(1, "acpi", "acpi event: pm timer");
	return 0;
}

/* The order of initialization comes from the ACPICA Programmer's Reference */
void main(int argc, char *argv[])
{
	ACPI_STATUS status;
	extern UINT32 AcpiDbgLevel;

	status = AcpiInitializeSubsystem();
	if (ACPI_FAILURE(status))
		fail("failed to init subsystem");
	status = AcpiInitializeTables(NULL, 4, 0);
	if (ACPI_FAILURE(status))
		fail("failed to init tables");
	status = AcpiLoadTables();
	if (ACPI_FAILURE(status))
		fail("failed to load tables");
	status = AcpiEnableSubsystem(ACPI_FULL_INITIALIZATION);
	if (ACPI_FAILURE(status))
		fail("failed to enable subsystem");
	/* ACPI Handlers should be installed at this time */
	status = AcpiInitializeObjects(ACPI_FULL_INITIALIZATION);
	if (ACPI_FAILURE(status))
		fail("failed to initialize objects");
	/* Other ACPI initialization can occur here including:
	 * 	Enumerating devices using _HID method
	 * 	Loading, configuring, installing device drivers
	 *	Device Driver install handlers for other address spaces
	 *	Enumerating PCI devices and loading PCIConfig handlers
	 */
	//AcpiDbgLevel |= ACPI_LV_VERBOSITY1 | ACPI_LV_FUNCTIONS;
	//status = AcpiInstallGlobalEventHandler(shutdownhandler, nil);
	status = AcpiInstallFixedEventHandler(ACPI_EVENT_POWER_BUTTON, shutdownhandler, nil);
	status = AcpiInstallFixedEventHandler(ACPI_EVENT_PMTIMER, timerhandler, nil);
	if (ACPI_FAILURE(status))
		fail("failed to install fixed event handler");
	print("Fixed Event Handler installed!\n");
	status = AcpiEnableEvent(ACPI_EVENT_POWER_BUTTON, 0);
	if (ACPI_FAILURE(status))
		fail("failed to enable power button");
	status = AcpiEnableEvent(ACPI_EVENT_PMTIMER, 0);
	if (ACPI_FAILURE(status))
		fail("failed to enable pmtimer");
	while (1)
		sleep(30*1000);
	status = AcpiTerminate();
        if (ACPI_FAILURE(status))
                fail("failed to terminate acpi");
        exits(nil);
}
```

### DIFF
--- a/sys/src/9/amd64/devacpi.c
+++ b/sys/src/9/amd64/devacpi.c
@@ -13,6 +13,7 @@
 #include "dat.h"
 #include "fns.h"
 #include "io.h"
+#include "../port/error.h"
 
 #include "apic.h"
 #include "acpi.h"
@@ -55,7 +56,24 @@ static Atable **atableindex;
 static Rsdp *rsd;
 Dev acpidevtab;
 
+typedef struct Acpiio {
+	uint32_t start;
+	uint32_t end;
+} Acpiio;
+
+#define NACPIPORTS	16
+static Acpiio acpiiospace[NACPIPORTS];
+
 static int32_t acpimemread(Chan *c, void *a, int32_t n, int64_t off);
+static int32_t acpiioread(Chan *c, void *a, int32_t n, int64_t off);
+static int32_t acpiiowrite(Chan *c, void *a, int32_t n, int64_t off);
+static int32_t acpiintrread(Chan *c, void *a, int32_t n, int64_t off);
+static int32_t acpiintrwrite(Chan *c, void *a, int32_t n, int64_t off);
+static void acpihandler(Ureg *u, void *v);
+static void *acpiintrvec;
+static Proc *intrproc;
+static int intrprocpid;
+static uint16_t pm1status;
 
 static char * devname(void)
 {
@@ -1901,7 +1919,15 @@ static void initgpes(void)
 
 static void acpiioalloc(unsigned int addr, int len)
 {
-	ioalloc(addr, len, 1, "ACPI");
+	Acpiio *io;
+
+	for (io = acpiiospace; io < acpiiospace + NACPIPORTS; io++) {
+		if (io->start != 0)
+			continue;
+		io->start = addr;
+		io->end = addr + len;
+		break;
+	}
 }
 
 static void acpiinitonce(void)
@@ -1910,12 +1936,14 @@ static void acpiinitonce(void)
 	if (root != nil)
 		print("ACPI initialized\n");
 	addarchfile("acpimem", 0444, acpimemread, nil);
+	/* XXX these permissions are too loose? */
+	addarchfile("acpiio", 0666, acpiioread, acpiiowrite);
+	addarchfile("acpiintr", 0666, acpiintrread, acpiintrwrite);
 	/*
 	 * should use fadt->xpm* and fadt->xgpe* registers for 64 bits.
 	 * We are not ready in this kernel for that.
 	 */
 	assert(fadt != nil);
-#if 0	/* These iomaps intefere with acpica library */
 	acpiioalloc(fadt->smicmd, 1);
 	acpiioalloc(fadt->pm1aevtblk, fadt->pm1evtlen);
 	acpiioalloc(fadt->pm1bevtblk, fadt->pm1evtlen);
@@ -1925,7 +1953,6 @@ static void acpiinitonce(void)
 	acpiioalloc(fadt->pmtmrblk, fadt->pmtmrlen);
 	acpiioalloc(fadt->gpe0blk, fadt->gpe0blklen);
 	acpiioalloc(fadt->gpe1blk, fadt->gpe1blklen);
-#endif
 
 	initgpes();
 #ifdef RON_SAYS_CONFIG_WE_ARE_NOT_WORTHY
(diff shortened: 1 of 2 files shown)


sys/src/9/amd64/devacpi.c
```diff
@@ -1910,12 +1936,14 @@ static void acpiinitonce(void)
 	if (root != nil)
 		print("ACPI initialized\n");
 	addarchfile("acpimem", 0444, acpimemread, nil);
+	/* XXX these permissions are too loose? */
+	addarchfile("acpiio", 0666, acpiioread, acpiiowrite);
```
rminnich: Once we really turn security up, these can be 660 and only usable by eve. But it will do for now. Good to leave the note in.

sys/src/9/amd64/devacpi.c
```diff
@@ -2066,6 +2093,189 @@ acpimemread(Chan *c, void *a, int32_t n, int64_t off)
...
+	char *p;
+
+	p = a;
+	if (p == nil || n < 1)
+		return 0;
+	if (*p == '1') {
+		acpiintrvec = intrenable(fadt->sciint, acpihandler, nil, BUSUNKNOWN,
+		    "ACPI interrupt handler");
+		if (acpiintrvec) {
+			intrproc = externup();
```
rminnich: What happens if the proc exits? Maybe you need to look in close and clear up these two variables.

sys/src/9/amd64/devacpi.c
```diff
@@ -2066,6 +2093,189 @@ acpimemread(Chan *c, void *a, int32_t n, int64_t off)
...
+acpihandler(Ureg *u, void *v)
+{
+	unsigned int status;
+
+	status = getpm1sts();
+	pm1status = status;
+	setpm1sts(status);
+	if (intrproc == nil)
+		return;
+	/* XXX there must be a better way to test if user Proc is okay */
```
rminnich: this is pretty dangerous as the proc may have exited and the pointer is pointing to ... what?

The common way to do this is use the chan as  the control point. As long as the Chan remains open, then intrproc is set. The file for the chan should be marked as exclusive, i.e. only gets one opener. If you set that up right this entire if block, as well as the test against pid, below, can be removed.

rminnich: oh and when the Chan is closed, intrproc is cleared. No locks needed for setting and clearing it, as the Chan and exclusive open ensure that you're the only person manipulating those variables!

sys/src/libacpi/harvey.c
```diff
@@ -436,22 +436,65 @@ acpihandler(void *_, void *arg)
...
 #endif
 
+static struct handler ihandler;
+static Rendez irendez;
+static QLock irlock;
+static int intrpid;
+
+/*
+ * The acpica service routines can't run in the note handler
+ * because of a floating point exception. Fixable?
```
rminnich: we should not assume it is. what you are doing with the rendezvous is fine BUT: you could use the plan 9 threads and channels model, and have  a process sleeping on a chan. The acpihandler would write useful messages down the chan, which will wake up the thread.

sys/src/9/amd64/devacpi.c
```diff
@@ -1910,12 +1936,14 @@ static void acpiinitonce(void)
 	if (root != nil)
 		print("ACPI initialized\n");
 	addarchfile("acpimem", 0444, acpimemread, nil);
+	/* XXX these permissions are too loose? */
+	addarchfile("acpiio", 0666, acpiioread, acpiiowrite);
+	addarchfile("acpiintr", 0666, acpiintrread, acpiintrwrite);
```
rminnich: Mark this as exclusion open, i.e. single open at a time. This will make lots of other stuff easier.

sys/src/libacpi/harvey.c
rminnich: I've cherry-picked these two commits and will be trying to get them ready in accordance with my review. This stuff is too good not to bring in.

sevki:
hey @sanderssj, do you have a timeframe for completing this PR or should we carry this for you?

sanderssj:
I'll see if I can have it done by Friday, otherwise feel free to take what
you need from it. Thanks, and sorry for the delays.

sevki:
Thanks @sanderssj. We really appreciate the help.

rminnich:
Ron took it over.

PR #378, opened 2016-12-18
boot: make boot able to user #P/cons if it is there

rminnich:
screenconsole does not work on riscv yet. But riscv provides
fails.

Also add a boot.ld script so riscv toolchain doesn't set up
boot incorrectly.


rminnich:
btw, this will break, I just have to figure out how to write it.

rminnich:
This is tested and boots. The earlier change, with a boot.ld, broke booting on x86. That's weird.

clang booting is still broken but it appears broke some time ago and nobody notice.

rminnich:
well, here we go ....

PR #384, opened 2016-12-21
Fixing bldy building with GCC

elbing:
For no clang fans. Avoiding PIC error in kernel mode.
Added option is not a conflict for clang.


elbing:
Oh I was just writing that right now. I tested with clang building my apex tests. But travis will give us the rest. If clang would'nt like the flag, it would crash. The problem is:
GCC -> -Wno-frame-address
Clang -> -mno-implicit-float
First is not in clang and second is not in gcc so...

elbing:
Kernel doesn't boot with clang. Abandoned.

PR #386, opened 2016-12-21
Somewhat working native QEMU 9p support

dmgolubovsky:
Sorry it took so long, the problem was indeed non-trivial: QEMU implementation of 9p indeed has a lot of problems, but it seems the solution exists.

For now I can mount 9p tags defined on the QEMU command line using a command like this:
```
mount -d '#9' /dev/null /mnt/xxx tag
```
It is important to use -d '#9' because #M cannot handle 9p2000.u, some on-the-fly correction needed to be done over the received data.

I tested large files transfer (using dd), shows better speed and more stable than 9p over virtio-serial.

Mount tags can be listed using /dev/mtags (via devcons).

Things not working well so far:

1. When I mount a host directory containing files other than regular (e. g. sockets) I get malformed state buffer. This should be fixable easy, I just gave it a lower priority. Not sure what to do with such files though; they cannot be used by Harvey anyway.

2. I cannot detect unmounting. I try to detect close on the mounted channel, but it does not seem to work.

How it all works: I created a specialized device #9 which piggybacks on top of devmnt, but reimplements few methods to properly handle 9p2000.u. For each mount, a "phantom" device is created whose devtab is not addressable, and it has only read and write methods implemented (more methods are just not needed). Among other things I made the list of mount-only devices (dcok) unified across sysfile.c and chan.c - now there are 3 devices: M, N, and 9.

This PR solves the problem of native 9p access to the host file shares, but it does not solve the problem of arbitrary 9p proxies running on the host. This was achievable using virtio-serial from my recent PR for virtio-serial: I recommend to review it again, maybe something can be thrown away, but such mechanism is definitely needed.

Thanks.


sys/src/9/port/chan.c
```diff
@@ -14,6 +14,8 @@
 #include	"fns.h"
 #include	"../port/error.h"
 
+int checkdc(int dc);
```
rminnich: at some point this can go in fns.h ... later.

PR #390, opened 2016-12-28
riscv: more riscv fixes

rminnich:
rc now works.


elbing:
Needs rebase.

sys/src/9/port/syscallfmt.c
```diff
@@ -315,6 +315,11 @@ syscallfmt(uint8_t what, int syscallno, Ar0 *ar0, uint64_t start, uint64_t stop,
 			errstr = up->syserrstr;
 		fmtprint(&fmt, " = %#p", ar0->v);
 		break;
+	case RFORK:
+		if(ar0->v == (void*)-1)
+			errstr = up->syserrstr;
+		fmtprint(&fmt, " = %d", ar0->v);
+		break;
```
elbing: Wasn't rfork here? I lost anything...

rminnich: it was only there for the entry case, but not the exit case.

I'm about to pull in the akaros strace code I did, which is a way better job than what we have now. This won't matter as much.

elbing:
Ok. Thanks to our bros in Akaros, tell them about.

PR #391, opened 2016-12-29
9p proxy over virtqueue

dmgolubovsky:
I previously proposed to add 9p support to the virtio-serial-pci driver (aka virtconsole) but this was not accepted. Here I am proposing a similar solution, but using an userspace proxy program. The program (9pvpxy) simply reads 9p messages from its std input and sends to the virtqueue, receiving buffers from the virtqueue and minimally massaging them to write 9p messages of proper length to its std output.

Example usage:

```
srv -e '9pvpxy /dev/virtcon/virtio-console-2' vc2
mount -c -n /srv/vc2 /mnt/xxx
```

The proxy was tested same way as the 9p virtio was tested (dd file read-write). It showed lower speeds and less stability than the 9p native virtio, but it allows to have arbitrary 9p servers on the host (ufs is just an example) while native 9p does not allow that (cannot be connected to a pipe, only to a FS device).

Hopefully this approach is more acceptable.

Thanks.

### DIFF
--- a/sys/src/9/port/devvcon.c
+++ b/sys/src/9/port/devvcon.c
@@ -59,7 +59,11 @@ rwcommon(Vqctl *d, void *va, int32_t n, int qidx)
 		error("virtcon: queue low");
 		return -1;
 	}
-	uint8_t buf[n];
+	uint8_t *buf = malloc(n);
+	if(buf==nil) {
+		error("devvcon: no memory to allocate the exchange buffer");
+		return -1;
+	}
 	if(qidx) {
 		memmove(buf, va, n);
 	}
@@ -73,6 +77,7 @@ rwcommon(Vqctl *d, void *va, int32_t n, int qidx)
 		memmove(va, buf, n);
 	}
 	reldescr(vq, 1, descr);
+	free(buf);
 	return (rc >= 0)?n:rc;
 }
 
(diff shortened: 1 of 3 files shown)


dmgolubovsky:
The massaging of received over a pci-serial virtqueue 9p messages was needed because devmnt gets confused in some cases if read returns message length more than actually received. At least Rversion had such problem (so I could not get past that in my early experiments). But with virtqueues we cannot tell how much of the received buffer was filled without knowing something about the message structure. Hence the need in either the driver or an userspace proxy to implement such massaging.

The merged PR #386 does enable native virtio 9p in Harvey kernel, so we have it as well (should work in general, maybe few minor fixes will be needed). What did not work with native 9p is to have an arbitrary program serving 9p over a pipe (socket) running on the host and visible to Harvey. This is because of the way 9p implemented in QEMU: it is very much tied to the Linux filesystem architecture on both ends. So it cannot work with a pipe on the host, only with a specific FSDEV configured inside QEMU. Though such approach gives better performance (just one virtqueue kick per 9p request/response hence less VM context switches). PR #391 enables Harvey to use arbitrary host program over a Unix socket (like ufs in my example). Downsides of this approach are: two virtqueue kicks per 9p request/response (one kick per message) hence worse performance; malfunctioning host 9p program may block the whole QEMU (and then even keyboard/mouse capture cannot be released easily, I had to switch to other console and kill QEMU from there when ufs hang). But these two options give a good range of features  balanced with their bugs ;)

I'd like to hear from the team members who use these new features.

Thanks.

elbing:
Tell us how to run this with qemu (qemu/kvm command line) and I'll update Harvey's Wiki with virtio info, and in future default qemu scripts. I think most of us wouldn't know where to start :-D.

dmgolubovsky:
I started a Wiki page for that.

dmgolubovsky:
The Wiki page has been written (at least covered all existing features, corrections welcome).

https://github.com/Harvey-OS/harvey/wiki/Using-Virtio-(QEMU)

elbing:
Fine! Thanks.

PR #394, opened 2016-12-31
bldy: make init9.c not relative

rminnich:
it won't build otherwise.


rminnich:
i think this is the wrong way to fox this bldy bug

sevki:
@rminnich if you don't mind posting this over to [bldy/build](https://github.com/bldy/build/issues/new) with the repro steps I'll give fixin it a go

PR #395, opened 2016-12-31
bldy: demand that certain things have been built or fail

rminnich:

### DIFF
--- a/util/continuous-build.sh
+++ b/util/continuous-build.sh
@@ -22,6 +22,7 @@ else
 			./util/scanscan
  	else
 		./util/bldy //.:kernel
+		ls -l $ARCH/bin/init
 
 	fi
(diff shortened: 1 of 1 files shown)


sevki:
bldy will fail if things are not built. why do we need this?

rminnich:
because it's not building it, or at least not putting anything in amd64/bin/init. I don't know why. 

So we've been thinking bldy works all this time when it does not.

rminnich:
Or at least I have. I went to set up bldy for riscv and suddenly realized it was not working for amd64 :-(

sevki:
it definitely does build it. are you looking into `build_out/amd64/bin` or `amd64/bin`?

sevki:
is this resolved?

PR #397, opened 2017-01-03
WIP: bldy: have bldy build more things

rminnich:
doesn't work yet, but it will


rminnich:
So this does it. The result is mostly a mirror in build_out of the tree we need to boot.
But it has to be moved.
Note that acpi is not building, as is acpi/irq.

rminnich:
API IS NOW building ... sorry

PR #398, opened 2017-01-03
bldy: change the name of the objcopy file

sevki:
Signed-off-by: Sevki <s@sevki.org>

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -1,16 +1,7 @@
-objcopy(
-	name="harvey.32bit",
-	deps=[
-		"//sys/src/9/amd64:harvey"
-	],
-	infile="elf64-x86-64",
-	outfile="elf32-i386",
-)
-
 move(
 	name="kernel",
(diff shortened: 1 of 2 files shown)


BUILD
```diff
@@ -86,7 +86,7 @@ move(
 		"amd64/lib/libthread.a": "lib/libthread.a",
 		"amd64/lib/libusb.a": "lib/libusb.a",
 		"amd64/lib/libventi.a": "lib/libventi.a",
-		"amd64/harvey": "bin/harvey",
+		"amd64/harvey.32bit": "bin/harvey.32bit",
```
rminnich: almost. You might want to just use mine; it puts it in sys/src/9/amd64/harvey.32bit, which is what all the scripts expect and which will work assuming union mounts.

rminnich:
I'm almost certain that we don't need this one now.

Issue #110, opened 2016-03-29
See if qremove is being called incorrectly 


rminnich:
That function was meant to be called with the queue locked.  #mnt doesn't
lock it - nor does it or any other file have a mechanism to do so.  So it's
always a bad idea to call qremove.  It looks like #mnt wants qget().
## Signed-off-by: Barret Rhoden brho@cs.berkeley.edu

Not sure if this is a problem in Plan 9 / Harvey or not.

 kern/drivers/dev/mnt.c | 2 +-
 kern/include/ns.h      | 1 -
 kern/src/ns/qio.c      | 2 +-
 3 files changed, 2 insertions(+), 3 deletions(-)

diff --git a/kern/drivers/dev/mnt.c b/kern/drivers/dev/mnt.c
index b9036cc438bb..df3dd304b319 100644
--- a/kern/drivers/dev/mnt.c
+++ b/kern/drivers/dev/mnt.c
@@ -969,7 +969,7 @@ int mntrpcread(struct mnt *m, struct mntrpc *r)
        l = &r->b;
        *l = NULL;
        do {
-               b = qremove(m->q);
-               b = qget(m->q);
              /\* TODO: have better block helpers for this and the memmove below _/
              b = linearizeblock(b);
              if (hlen > 0) {
  diff --git a/kern/include/ns.h b/kern/include/ns.h
  index 53d77b5a0f83..32e3052695eb 100644
  --- a/kern/include/ns.h
  +++ b/kern/include/ns.h
  @@ -836,7 +836,6 @@ ssize_t qpass(struct queue *, struct block *);
  ssize_t qpassnolim(struct queue *, struct block *);
  void qputback(struct queue *, struct block *);
  long qread(struct queue *, void *, int);
  -struct block *qremove(struct queue *);
  void qreopen(struct queue *);
  void qsetlimit(struct queue *, int);
  int qwindow(struct queue *);
  diff --git a/kern/src/ns/qio.c b/kern/src/ns/qio.c
  index d6ddd278c96e..066a4f3e379e 100644
  --- a/kern/src/ns/qio.c
  +++ b/kern/src/ns/qio.c
  @@ -1177,7 +1177,7 @@ void qaddlist(struct queue *q, struct block *b)
  /_
  -  called with q ilocked
    */
    -struct block *qremove(struct queue *q)
    +static struct block *qremove(struct queue *q)
    {
      struct block *b;
## 

2.8.0.rc3.226.g39d4020

[...]

PR #399, opened 2017-01-05
Fix for issue #110

keedon:
I've changed qremove in devmnt to qget to ensure lock is held before manipulating the queue

Also:
* I've added some compilation flags to the build to make gdb work correctly
* there is a new start up script GO9PCPUD which will halt at start of boot to allow gdb to connect


amd64/include/cflags.json
```diff
@@ -20,7 +20,9 @@
 				"-Wno-main"
 			],
 			"/usr/bin/gcc": [
-				"-Wno-frame-address"
+				"-Wno-frame-address",
+				"-fvar-tracking",
+				"-fvar-tracking-assignments"
```
rminnich: bless you. I could NOT figure this out.

sys/src/regress/floatnotify.c
```diff
@@ -0,0 +1,27 @@
+#include <u.h>
```
rminnich: can you put a message in here about what this does, and a copyright notice? I don't understand what you're doing here.

sys/src/9/port/devmnt.c
```diff
@@ -963,7 +963,7 @@ mntrpcread(Mnt *mnt, Mntrpc *r)
 	l = &r->b;
 	*l = nil;
 	do {
-		b = qremove(mnt->q);
+		b = qget(mnt->q);
```
rminnich: I betcha this is my fault from when I split up the devmnt IO. It sure seems right. We'll see.

I would prefer it if you could split this one change out into its own PR. The other changes are not really related.

PR #404, opened 2017-01-13
Add usb ports to Harvey GO9PTERM script

aahud:
This provides the template to add specific usb devices. If preferred I should be able to simply have Harvey take control over all usb devices. On the system, but I can't really think of a middle ground, because I wouldn't be able to specify usb devices for each user.


util/GO9PTERM
```diff
@@ -16,8 +16,11 @@ if [ "$(uname)" = "Linux" ] && [ -e /dev/kvm ]; then
 	fi
 fi
 
+# specific usb devices added by '-device usb-host,hostbus=X,hostaddr=X'
+# replacing 'X' with the device numbers provided by a tool like 'lsusb'
 read -r cmd <<EOF
 $kvmdo qemu-system-x86_64 -s -cpu Opteron_G1 -smp 2 -m 2048 $kvmflag \
+-usb \
```
rminnich: could you put some examples of usb usage in here? I don't understand this.
Maybe you could how you add your mouse by finding it with lsusb and then adding it to the command line or something

aahud: Line 19 is the usb device example. You want lsusb output examples in the file as well? There is no guarantee what address a given usb device will receive from what I can tell.
I can't put comments in the qemu command because I'm not sure how to escape the backslashes.

elbing:
You need to sign your commit, git commit -s --amend will do the tirck, just git push -f foo aahud to your fork will upload it.

util/GO9PTERM
rminnich: ok, how about you show the result of executing those commands and then what you add to the qemu command?

I'll try this PR now.

util/GO9PTERM
```diff
@@ -16,8 +16,14 @@ if [ "$(uname)" = "Linux" ] && [ -e /dev/kvm ]; then
 	fi
 fi
 
+# The -usb flag is necessary for the system to recognize usb input
```
rminnich: instead of input, say 'devices'. It's not just input.
For people who don't know much about lsusb, why not show a sample command and where 'X' comes from? You say 'X' two times below in your example; is it the same X or a different X?

So show an an example of an lsub command,then show what hostbus=X and hostaddr=Y would be in the command below, i.e., show a sample command.

rminnich: by the way this level of discussion in a review can be pretty common, especially for your first couple PRs, so do not get discouraged.

sevki:
hey @aahud, do you have a time from for finishing this? If you are too busy to finish we can ask for someone to take it over and finish it for you :smile:

aahud:
Branch is updated now. I believe that I incorporated all of the suggestions given. My absence has been more related to the start of my semester than anything else, and for that I apologize. There will be a lot more availability from me now.

sevki:
thanks @aahud. can you please sign your last commit, we can't merge without it

PR #408, opened 2017-01-30
Fix warnings in screenlock

gmacd:
Screenlock now looks fine if the image can't be loaded for some reason.
Tested manually.


sys/src/cmd/screenlock.c
```diff
@@ -241,21 +240,23 @@ lockscreen(void)
 	procrfork(blanker, nil, 4096, RFFDG);
 	fd = open(pic, OREAD);
 	if(fd > 0){
+		draw(screen, screen->r, display->black, nil, ZP);
```
rminnich: Do you really want to draw a black box here before you know the readimage succeeded?

rminnich: OH got it! yeah you do!

rminnich: oh wait then should you draw the blank even if the open fails?

gmacd: If the open fails, I still want a black screen, because the text is white.  As it was, it would be white on white which isn't very helpful.

rminnich: I would argue that the blank should be even before the open, but outside the if. As it is it looks like your blanking is dependent on the if. I may still be misreading it.

gmacd: You're right.  I'll take a look at it tomorrow.

elbing:
Please, push -f for forcing travis to rebuild.

sevki:
@elbing  you can trigger builds by going to `https://api.travis-ci.org/build/{ID}/restart` for example for this build `https://travis-ci.org/Harvey-OS/harvey/builds/196765952` id would be `196765952` or you can just click the restart build button to do it.

But this pr requires a rebase not a rebuild :smiley:

sevki:
I stand corrected it needed a rebuild. but it still needs a rebase :smiley:

gmacd:
I'm closing this and restarting - I've screwed up git - still getting the hang of using it like this.

PR #412, opened 2017-01-31
Fix warning in screenlock, handle image not found better

gmacd:
Handle case where image isn't found, and tidy code a bit.

### DIFF
--- a/sys/src/cmd/screenlock.c
+++ b/sys/src/cmd/screenlock.c
@@ -211,27 +211,23 @@ void
 lockscreen(void)
 {
 	enum { Nfld = 5, Fldlen = 12, Cursorlen = 2*4 + 2*2*16, };
-	char *s;
-	char buf[Nfld*Fldlen], *flds[Nfld], newcmd[128], cbuf[Cursorlen];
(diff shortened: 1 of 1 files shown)


sys/src/cmd/screenlock.c
```diff
@@ -211,62 +211,63 @@ void
...
+	if(initdraw(nil, nil, "screenlock") < 0)
 		sysfatal("initdraw failed");
 	if(display == nil)
 		error("no display");
 
 	/* screen is now open and covered.  grab mouse and hold on tight */
 	procrfork(grabmouse, nil, 4096, RFFDG);
 	procrfork(blanker, nil, 4096, RFFDG);
+
+	draw(screen, screen->r, display->black, nil, ZP);
```
rminnich: You could swap these two lines and do draw(screen, r, etc. etc.)
Not a big deal but ...

gmacd:
Why not?  I need the git practice :)

elbing:
Ok @gmacd, can you merge it or no permissions?

gmacd:
I don't think I have permission

PR #413, opened 2017-02-01
Bring back ^T

rminnich:
Very handy. Only thing is I'm not sure where to hook in catchcontrolt.
It has to be in non-interrupt mode. But one of you can work it out.


rminnich:
This change has been overtaken by events, namely, we're going back to kernel consoles.

PR #418, opened 2017-02-03
Fix some warnings ACPI code

gmacd:
Fixes all but one ACPI warning.


sys/src/libacpi/acfileio.c
```diff
@@ -125,6 +125,7 @@
 
 /* Local prototypes */
 
+#if 0
```
rminnich: delete this function. My bad. But we try to avoid #if 0 in harvey code. Thanks!

sys/src/libacpi/harvey.c
```diff
@@ -195,11 +190,9 @@ AcpiOsReadPciConfiguration(ACPI_PCI_ID * PciId,
 }
 
 ACPI_STATUS
-AcpiOsWritePciConfiguration(ACPI_PCI_ID * PciId,
-							UINT32 Reg, UINT64 Value, UINT32 Width)
+AcpiOsWritePciConfiguration(ACPI_PCI_ID * PciId, UINT32 Reg, UINT64 Value, UINT32 Width)
```
rminnich: I like this. I prefer one line and the plan 9 style is not to break on 80 columns. Thanks.

sys/src/libacpi/olsneracpi.c
```diff
@@ -234,6 +234,7 @@ ACPI_STATUS FindIOAPICs(int *pic_mode) {
 	return AE_OK;
 }
 
+#if 0
```
rminnich: delete this one too for now.

sys/src/libacpi/acfileio.c
gmacd: This file consists mostly of functions with their contents #if 0'd.  Are you suggesting getting rid of the function and the prototypes?

sys/src/libacpi/olsneracpi.c
gmacd: Ok - there are other #if 0'd functions in here - should these be removed too?

rminnich: They should be removed,  but not as part of this PR. The rule is you keep PRs small and focused. Makes it easier to revert them later.

elbing:
My fault. I told him in slack about removing dead code.

gmacd:
Ok, I've fought with git, and removed these functions now.  With a little luck, it'll even build.

PR #420, opened 2017-02-05
Fix rune warnings in acme

gmacd:
Having to cast these string literals to (Rune*) doesn't look great, but it does remove the warnings for now, as previously mentione in #14.

### DIFF
--- a/sys/src/cmd/acme/cols.c
+++ b/sys/src/cmd/acme/cols.c
@@ -40,7 +40,7 @@ colinit(Column *c, Rectangle r)
 	r1.min.y = r1.max.y;
 	r1.max.y += Border;
 	draw(screen, r1, display->black, nil, ZP);
-	textinsert(t, 0, L"New Cut Paste Snarf Sort Zerox Delcol ", 38, TRUE);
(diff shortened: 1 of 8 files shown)


sevki:
@gmacd what are you using for formatting?

gmacd:
@sevki In this case, I formatted by hand, since I was looking at that area.  I tried clang-format a few days ago, but didn't like what it did in some cases (I think it was bunching up multiple enums per line, where there had been 1 per line before).

rminnich:
So, while we might not always like what clang-format does, the rule (I think) should be: that's the way it goes. I really want to get away from all this manual formatting nonsense. I'd really be happy  if we all just accepted what clang-format does. Can you make a rule that for clang-format that does what you want? Can we get a better formatter?

gmacd:
That's fine. I even accepted 8 space tabs easily enough :) I'll take a look at the clang format settings tonight - originally I wasn't sure if using it was enforced or suggested.

rminnich:
it's strongly suggested by me. I'd like to have coreboot rules, which are linux rules but allow // comments. But let's see what alvaro says.

elbing:
Fine for me.

PR #423, opened 2017-02-07
Change IRQRouteData pin to UINT64

gmacd:
This fixes the last warning in the ACPI code.


rminnich:
be aware that changes to the irq command need to be tested (I just did) on real machines or vmware. Can't be tested in qemu.

PR #428, opened 2017-02-11
Break apart malloc interface and implementation

gmacd:
This should make it easier to replace the allocator implementation, and also to test and analyze.

### DIFF
--- a/sys/src/9/amd64/qmalloc.c
+++ b/sys/src/9/amd64/qmalloc.c
@@ -20,7 +20,8 @@
 #include "mem.h"
 #include "dat.h"
 #include "fns.h"
-#include	<pool.h>
+#include <pool.h>
+#include "malloc.h"
 
 typedef double Align;
 typedef union Header Header;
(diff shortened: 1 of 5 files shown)


gmacd:
I didn't move anything into amd64 that wasn't there already.  The malloc.c file contains functions that were in qmalloc.c.  Perhaps I'm looking at the wrong area?  Is this not the allocator you suggested replacing?

rminnich:
yeah you're right. OK, I guess this is fine at least for measurement. The 'always zero' behavior is something we did early on and I don't want to change it without testing.

Did you boot and test with this yet?

gmacd:
I've been testing in qemu.  So far seems fine.  If the intended behaviour is to always clear for both malloc and mallocz I can add that behaviour back in - I removed it because it looked like a typo.

rminnich:
I'd feel a wee bit happier if we had a solid regression test first. malloc changes can result in very weird bugs. Could you do a test? See the regression test device which should let you poke sticks at malloc in all kinds of ways.

gmacd:
I'd feel happier too - I'm looking at tests ATM, so it's good to get some feedback before going further down the path.

sys/src/9/amd64/qmalloc.c
```diff
@@ -665,19 +655,22 @@ mallocinit(void)
...
-	 * First (simple) cut.
-	 * Pump it up when you don't really need it.
-	 * Pump it up until you can feel it.
-	 */
-	if(nunits < NUNITS(128*KiB))
-		nunits = NUNITS(128*KiB);
-	if(nunits > tailnunits)
-		nunits = tailnunits;
-	tailnunits -= nunits;
+static struct Allocator _qmallocAllocator = {
```
rminnich: I don't see the point of this. You're putting an indirection on functions right in the middle of a critical piece of code. This means we may not be able to use clang's ability to inline at link time. For what purpose?

gmacd: The main reason was to separate out the core allocator code.  Since then I've been writing code to record all the allocator events so that I can analyze them and play them back in tests and benchmarks.  Breaking up the code a bit seemed a good way to enable this code to be reused when looking at implementations.

Also, after looking at the differences between the amd64 and riscv implementations, I noticed that riscv uses qmallocalign for standard qmalloc.  It seemed that with the Allocator struct I could move all the allocator code into port, and leave the Allocator definition in amd64 and riscv, with the riscv definition having qmalloc and qmallocalign pointing at the same function.

Looking at it again, something similar could be achieved without the indirection by just calling the functions in qmalloc (or some other linked implementation) directly, although not quite sure how best to handle the riscv case without duplicating code, and avoiding using the preprocessor.

gmacd: All that being said, is it better to close this PR, since it clearly doesn't add any value yet, or to iterate on it?

PR #426, opened 2017-02-09
Bringing back console to kernel

elbing:
Round 1

Signed-off-by: Álvaro Jurado <elbingmiss@gmail.com>

PR #430, opened 2017-02-11
i8042.c: don't push the char twice

rminnich:
It's not right for the kernel-based
console I think, since you interpret the char
below. With this one change in alvaros revertcons3 branch
things work. But this probably breaks the external console
so don't push it.

don't push this, I only tested it against PR 426 not
against upstream


rminnich:
confirmed this breaks upstream and ONLY works when 426 is applied.

elbing:
Included in #426

PR #434, opened 2017-02-13
Bringing back insecure telnet for development purposes

elbing:
Now with console in kernel.


rminnich:
I suggest you keep this as your own private patch. It's a massive security hole and I don't think it should be turned back on.

elbing:
Has sense. Bye telnet.

PR #439, opened 2017-02-22
build: get rid of sh dependencies

rminnich:
and clean up heinous shell code while we're at it.

Temporarily disable the syscall command. It has obsolete dependencies
on 32-bit ABIs and kenc issues. Someone can fix it someday if they care.

If someone can fix that awful fossil conf.rc mess I'd be grateful.
I'm done for the day.


elbing:
Oh man... That's there since early days. I'll try adding it to mksys.

rminnich:
I have this almost working but I'm going to crash now. Kids made me sick.

rminnich:
actually I would not worry about that fossil and venti conf.rc stuff. We want those file systems to go away anyway.

elbing:
I could evaluate how of all of that parsed script is still needed and for what. Though I'm back to apex.
Have fun with kids, grandpa.

PR #440, opened 2017-02-22
cflags: remove ToolOpts and just use include

rminnich:
I should have done it this way in the first place.


elbing:
Seems we learned something at the end of all of this. Ok, needs rebase.

PR #444, opened 2017-02-23
Fix remaining problems for build on mac concerning sh

gmacd:
Typical build no longer needs sh, so remove from tool dependencies.  Defaults to 'sh', but can be overridden with SHELL environment variable.  This gets us closer to being able to build on mac with gcc.


rminnich:
Confirmed this build works and boots on linux.

PR #448, opened 2017-02-25
Bind devrtc to /dev/rtc by default

gmacd:


gmacd:
You're right, it should have been #r.  I tested using GO9PCPU, where it works, but when I tried on GO9PTERM it failed.  It seems that devrtc is being added to the namespace for CPU but not for TERM.  So there's a little bit more to it than I thought.

gmacd:
I need to look into this a bit more.  Closing for now.

PR #450, opened 2017-02-26
Add binding for /dev/rtc

gmacd:
Ensures /dev/rtc is available for both cpu and terminal.


rminnich:
but will this interfere with the binding in cpurc? You ought to check.

gmacd:
It didn't seem to, it just got bound earlier (it was higher up the list
shown by ns).

PR #458, opened 2017-03-03
Troff

floren:
Trying it again, with signed-off this time.

Run BUILDME and it will build. I don't know how to translate that into a .json format so let's consider it a WIP.


elbing:
John, please, move it to APEX. Here place the PR for removing it. Same for rest of programs with APEX dependency.

PR #459, opened 2017-03-04
two new build features on request: globbing and dependency checking

rminnich:
see sys/src/cmds/build.json and sys/src/libacpi for examples.

Globbing simplifies things a lot.

The dependency checking is pretty simple but if you run build with -D
it will only build .c files needed to build the target. It's not nearly
as good as otehr tools (bldy) but it's "good enough" to make things
a lot faster.

I suggest for now that the first time you run build you "prime the pump"
with a non -D build. After that it will mostly work.

YMMV. Please remember, build is a script written in Go that's designed
to build harvey. It's not a general purpose build tool.


rminnich:
oh boy, let's pull this trigger!

PR #462, opened 2017-03-04
bldy: let bldy take a vacation for a bit

rminnich:
Bldy is a brilliant prototype but we want to rethink some aspects
of its implementation. Until we're sure how it's going to work
out, take it out of the test matrix (it's pushing our travis
times to 30 minutes) and don't get it or mention it in
bootstrap.sh. Do this in a way that lets us bring it back
when we're ready.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,6 @@ env:
   - PATH=$PATH:$HARVEY/util
   matrix:
   - OLD_BUILD=true
-  - OLD_BUILD=false
 addons:
   coverity_scan:
     project:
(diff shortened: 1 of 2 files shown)


sevki:
can we try instead
```
matrix:
   - OLD_BUILD=true
   - OLD_BUILD=false
  allow_failures:
    - OLD_BUILD: false
```

rminnich:
yes but are you ok if we take the bldy install out of bootstrap.sh? My original hope was to transition to bldy but we're just not ready. It's confusing people to have two build tools. This means comment out two lines.

PR #465, opened 2017-03-07
Added TLS 1.1 and 1.2

fuchicar:
Added the necessary library files to use tls 1.2


sys/src/libmp/port/cnfield.c
```diff
@@ -0,0 +1,123 @@
+/*
```
rminnich: This file is new? Where's it from ? What's the license? Same question for all these files.

sys/include/libsec.h
```diff
@@ -7,8 +7,6 @@
  * in the LICENSE file.
  */
 
-
-
 #ifndef _MPINT
```
rminnich: No. We don't do this in harvey, we follow the Plan 9 standards. This #ifdef has to go.

sys/src/libsec/port/dh.c
```diff
@@ -0,0 +1,83 @@
+/*
+ * This file is part of the UCB release of Plan 9. It is subject to the license
+ * terms in the LICENSE file found in the top-level directory of this
+ * distribution and at http://akaros.cs.berkeley.edu/files/Plan9License. No
+ * part of the UCB release of Plan 9, including this file, may be copied,
+ * modified, propagated, or distributed except according to the terms contained
+ * in the LICENSE file.
+ */
+
+#include "os.h"
```
rminnich: What's os.h? Is it a .h that includes .h? Then it needs to not do that. .h do not include .h in our code base. I realize that rule is broken in places but ... let's try to get this right.

PR #466, opened 2017-03-08
add freetype library & ttf2subf so we can import ttf fonts.

floren:
TODO: get a modern version of freetype and port p9p's fontsrv


floren:
PTAL, just deleted files

sevki:
even better :)

PR #467, opened 2017-03-08
Add Go fonts. Set harvey to use Go Mono 14pt by default.

floren:


floren:
As requested.

PR #469, opened 2017-03-08
stats: get default font

elbing:

### DIFF
--- a/sys/src/cmd/stats.c
+++ b/sys/src/cmd/stats.c
@@ -277,7 +277,8 @@ mkcol(int i, int c0, int c1, int c2)
 void
 colinit(void)
 {
-	mediumfont = openfont(display, "/lib/font/bit/pelm/latin1.8.font");
+
+	mediumfont = openfont(display, getenv("font"));
 	if(mediumfont == nil)
 		mediumfont = font;
 


sys/src/cmd/stats.c
```diff
@@ -277,9 +277,12 @@ mkcol(int i, int c0, int c1, int c2)
 void
 colinit(void)
 {
-	mediumfont = openfont(display, "/lib/font/bit/pelm/latin1.8.font");
+
```
rminnich: This is kinda gross. We have to do this to *every* *program* that wants a font that doesn't suck?

There has to be a better way.

sys/src/cmd/stats.c
```diff
@@ -277,9 +277,12 @@ mkcol(int i, int c0, int c1, int c2)
...
+	mediumfont = openfont(display, getenv("font"));
```
rminnich: Could we have a library function that just does this instead?

sys/src/cmd/stats.c
```diff
@@ -277,9 +277,12 @@ mkcol(int i, int c0, int c1, int c2)
...
 	if(mediumfont == nil)
 		mediumfont = font;
+	else
+		mediumfont = openfont(display, "/lib/font/bit/go/Go-Mono/Go-Mono.14.font");
```
floren: We already set $font to Go-Mono.14.font in /usr/harvey/lib/profile, so that's what it should be getting. I think you should be doing this instead:

mediumfont = openfont(display, getenv("font"));
if (mediumfont == nil)
   mediumfont = openfont(display, "/lib/font/bit/pelm/latin1.8.font");
if (mediumfont == nil)
   mediumfont = font;

But yeah we should make this a library function, "opendefaultfont" or whatever that attempts to first open $font and then whatever the fallback is (pelm)

elbing:
#472

PR #470, opened 2017-03-08
fix troff fonts

floren:
Send fixcase to hell.


gmacd:
A bit late, but looks like this PR causes problems on Mac, where the filesystem is case insensitive by default.  Now, if I clone a fresh repo on my mac, it thinks the HB, HI and HX files have been modified.

This is going to be quite off-putting to anyone trying to develop on a mac.

elbing:
Add them to .gitignore.

gmacd:
At first attempt, that doesn't work.

rminnich:
we made a decision a while back to not work on case-insensitive file systems. Is there any way to remove or change the conflicting names so they work on case-insensitive file systems without needing fixcase? I have no idea.

PR #473, opened 2017-03-09
Winwatch meets default font

elbing:

### DIFF
--- a/sys/src/cmd/winwatch.c
+++ b/sys/src/cmd/winwatch.c
@@ -256,7 +256,6 @@ main(int argc, char **argv)
 	int Etimer;
 	Event e;
 
-	fontname = "/lib/font/bit/lucidasans/unicode.8.font";
 	ARGBEGIN{
 	case 'f':
 		fontname = EARGF(usage());
@@ -277,8 +276,10 @@ main(int argc, char **argv)
 	lightblue = allocimagemix(display, DPalebluegreen, DWhite);
 	if(lightblue == nil)
 		sysfatal("allocimagemix: %r");
-	if((font = openfont(display, fontname)) == nil)
-		sysfatal("font '%s' not found", fontname);
+	if(fontname == nil)
+		font = opendefaultfont(display);
+	else
+		font = openfont(display, fontname);
 
 	refreshwin();
 	redraw(screen, 1);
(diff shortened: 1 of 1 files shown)


sys/src/cmd/winwatch.c
```diff
@@ -277,8 +276,11 @@ main(int argc, char **argv)
 	lightblue = allocimagemix(display, DPalebluegreen, DWhite);
 	if(lightblue == nil)
 		sysfatal("allocimagemix: %r");
-	if((font = openfont(display, fontname)) == nil)
-		sysfatal("font '%s' not found", fontname);
+	if(fontname == nil)
+		font = opendefaultfont(display);
```
rminnich: I still don't get why you don't make opendefaultfont take an optional name. You could reduce these lines to one call 
font = opendefaultfont(display, fontname);
done

elbing:
Then adding a check to original openfont would be the same. Also if it opens default, shouldn't be opening another, isn't it?
The question I think is "why font is a flag"? If you set it up by default in profile, why would you change for one app? Maybe it had sense 25 years ago but I can't see the advantage, usually it could be a sets of default fonts in profile for title bars, window terminals, etc...

sys/src/cmd/winwatch.c
```diff
@@ -277,8 +276,11 @@ main(int argc, char **argv)
...
+	else
+		if((font = openfont(display, fontname)) == nil)
```
floren: I think you still need to sysfatal after this.

elbing: Yep, in case font wouldn't exist winwatch just wouldn't start. Ok, later.

elbing: sysfatal is already there, Ron just commented the line before.

sys/src/cmd/winwatch.c
elbing: Because that solves this particular case only. In programs in what there's no any -f font flag, we should include always a nil char *name just for that. Or a font flag. The main purpose of opendefaultfont func was avoiding the cases of not using the default, no any font specified or it's harcoded. A name second argument would be duplicating the openfont func and just for this specifical case in this program.
The simpler solution for winwatch shouldn't be one line for check the font, but removing font flag at all. Because seeing the purpose of the app, choosing font hasn't any sense here.
Agree?

rminnich:
is this ready?

PR #474, opened 2017-03-09
Added TLS 1.1 and 1.2

fuchicar:
Added the necessary library files to use tls 1.2

@rminnich about ifndef in libsec: It came from plan 9 code, I didn't introduce it. If you want I can remove it, but I think it would be better to do it in a separate PR because it means I'll change a lot of files not only under libsec but in cmd tree as well. Do you agree?


rminnich:
yeah, you're right, it must have been when they imported it in the first place.

elbing:
Be carefull with cmd stuff. All which needs pcc muat be moved to apex/cmd or replaced with newer ports. Many ifdefs comes from ported programs and they were in tree as old ape.

bldybuild:
Commit f10806c1ba2d23b0ea934c3685d317a0c1951b17 is not signed-off.

Please fix these issues:

>	1. Rafa has not signed-off f10806c.

PR #475, opened 2017-03-09
Fix acme suicide.

fuchicar:
Fixed a problem that provokes a crash in acme when you reach at the end of a document and press the key down.


fuchicar:
@sevki sorry, the indentation was made by Atomic automatically. In Atomic it looked fine, but when I pushed it to github the result was a mess.

PR #477, opened 2017-03-09
make $font the default font for acme

floren:


elbing:
It could use opendefaultfont, but... then our acme would be incompatible with rest of Plan 9 derivatives, so... LGTM.

PR #483, opened 2017-03-11
acpiintr changes, partly to stop it flooding mcdonag's machine with prints

rminnich:


elbing:
The other chance is disabling acpiistart through bootargs as BSD or Linux, being enabled by default. For saying that "try disabling acpi" when someone had a problem.

rminnich:
no, acpistart will be enabled by control writes to /dev/acpictl. We don't need silly things in bootargs.

rminnich:
This is not enabling/disabling acpi, just acpi interrupts. These should only be enabled when the acpi daemon (not written) starts.

PR #485, opened 2017-03-13
Useless namespace

elbing:

### DIFF
--- a/lib/namespace
+++ b/lib/namespace
@@ -19,9 +19,6 @@ bind -b '#$' /dev
 bind -a 'α' /dev
 bind -a '#r' /dev
 
-# screen console
-mount -b /srv/screenconsole /dev
-
 # mount points
 mount -a /srv/slashn /n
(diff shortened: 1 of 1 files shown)


sys/src/cmd/acid/builtin.c
```diff
@@ -136,7 +136,6 @@ dosysr1(Node *r, Node*_)
 	r->op = OCONST;
 	r->type = TINT;
 	r->fmt = 'D';
-	fprint(2, "Thank Giacomo -- no sysr1 -- we'll bring it back\n");
```
rminnich: either verify that sysr1 is working again (it should be) or put an error and panic here. And, this should be a different PR.

rminnich: you can call sysr0 here, I brought it back a few months ago.

elbing: Restored.

PR #486, opened 2017-03-14
The syscall is sysr0

elbing:
(rest of the world is wrong)

### DIFF
--- a/sys/src/cmd/acid/builtin.c
+++ b/sys/src/cmd/acid/builtin.c
@@ -85,7 +85,7 @@ struct Btab
 	"status",	status,
 	"stop",		stop,
 	"strace",	strace,
-	"sysr1",	dosysr1,
+	"sysr0",	dosysr0,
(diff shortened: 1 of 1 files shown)


sys/src/cmd/acid/builtin.c
```diff
@@ -129,14 +129,14 @@ installbuiltin(void)
...
 {
-	extern int sysr1(void);
-	
+	extern int sysr0(void);
+
 	r->op = OCONST;
 	r->type = TINT;
 	r->fmt = 'D';
-	fprint(2, "Thank Giacomo -- no sysr1 -- we'll bring it back\n");
+	fprint(2, "Thank Giacomo -- no sysr0 -- we'll bring it back\n");
```
rminnich: I don't get it. You can call sysr0 so why not call it?

elbing: ah!

PR #487, opened 2017-03-14
Changed some scripts to use /usr/harvey instead of /usr/glenda

fuchicar:


elbing:
Fine! seems someone forgot a couple of things during 2015 summer days... ok, test if it's works restoring in harvey's lib/profile, rio -s -i riostart. Better than a gray screen. "-s" was not in original but I think most of us expects autoscrolling in term windows.

Issue #496, opened 2017-03-19
gdbserver must implement vMustReplyEmpty


rminnich:
vMustReplyEmpty is not supported and appears to be the reason gdbserver now dies.

gmacd:
This seems to be resolved now.

PR #498, opened 2017-03-19
Improve handlign of unsupported commands

gmacd:
Fixes issue #496 - handling of vMustReplyEmpty.

This is a very specific fix, can improve code in a later PR.

### DIFF
--- a/sys/src/cmd/gdbserver/gdbstub.c
+++ b/sys/src/cmd/gdbserver/gdbstub.c
@@ -946,17 +946,21 @@ gdb_serial_stub(struct state *ks, int port)
 				/* Fall through to default processing */
 			default:
 default_handle:
-				/*
-				 * I have no idea what to do.
-				 * Leave cmd processing on error, detach,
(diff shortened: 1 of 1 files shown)


gmacd:
According to [docs](https://sourceware.org/gdb/onlinedocs/gdb/Overview.html#Overview) any unsupported command should return "", so this can be generalized.
Further improvements that could be made here are to get rid of the error variable, which doesn't seem to do much, and look again at the code after the default_handle label.

sys/src/cmd/gdbserver/gdbstub.c
```diff
@@ -1083,6 +1087,10 @@ static struct state ks;
 void
 main(int argc, char **argv)
 {
+	syslog(0, "hello graham\n", 0);
```
rminnich: well, let's let this go in for now but .... graham?

gmacd: Haha, oops - let me fix that.

PR #503, opened 2017-03-23
Avoiding gcc warnings of 'if' indentation.

elbing:


elbing:
Maybe not with clang, but it's annoying.

PR #505, opened 2017-03-24
gdb: don't remove 9syscall source files, gdb gets confused.

rminnich:

### DIFF
--- a/sys/src/libc/libc.json
+++ b/sys/src/libc/libc.json
@@ -10,9 +10,6 @@
 	],
 	"Install": "/$ARCH/lib/",
 	"Library": "libc.a",
-	"Post": [
-		"rm -f 9syscall/*.s"
-	],
 	"Projects": [
 		"9syscall/9syscall.json"
 	],


rminnich:
macdonag see if this makes the .s files for gdb visible now.

elbing:
argh, this is crappy. Glibc or musl generate syscalls and they're not available for gdb. But well, for now... one day we'll have full gdb and maybe this won't be needed if it solves this.

rminnich:
I'm not sure I understand the problem. This won't be committed to the repo. We just won't remove the .s files when we build the library. They're in the .gitignore. What's the problem?

elbing:
The problem is Ron saying again "build doesn't clean properly the tree". So if you don't see a problem, I don't too :-).

rminnich:
so can someone check this with gdb and if it works approve it.

gmacd:
I can look at it tonight, UK time.

gmacd:
So this keeps the generated .s files around, which is a good thing - less confusing than deleting them immediately after build.  Unfortunately I still get the error from gdb:
Remote debugging using 192.168.0.81:1666
pread () at pread.s:7
7	pread.s: No such file or directory.

PR #511, opened 2017-03-29
Only send the main pid to gdbserver on qfThreadInfo

gmacd:
If we eventually want to support multiple processes, can re-add.


elbing:
Testers!

PR #524, opened 2017-04-18
Cleaned up warnings in libfreetype

dancrossnyc:
Mostly unused variables, or set-but-unused.

The unfortunate practice seems to be to define a variable,
initialize it to zero, and set it to one later.  Then, presumably,
someone could look at it with a debugger or something and see
that it had been set at some point (or not).  Perhaps this is
used as a rough signal to see that something has happened in a
function.

Unfortunately, an optimizing compiler will probably note that
the variable is set but unused and just remove it.  So we just
remove some of them ourselves.  Or whatever.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,16 +7,24 @@
 386/pbslba
 386/pbsraw
 
+aarch64/9*
+aarch64/bin
+aarch64/init
+aarch64/lib
+
 amd64/9*
 amd64/bin
 amd64/init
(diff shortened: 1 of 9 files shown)


bldybuild:
Commit 51da0deaf9e2eab5549ef8fd376f518f26e1fb0b is not signed-off.

Please fix these issues:

>	1. Dan Cross has not signed-off 51da0de.

PR #528, opened 2017-04-19
Stop hiding warnings.

dancrossnyc:
Reveal more in the way of build warnings.  We should just fix
'em instead of turning 'em off.


dancrossnyc:
(Deleting for now; it's broken.)

rminnich:
yeah those warnings turned out to be WAY harder to kill than we thought :-(

dancrossnyc:
I'll get 'em.

PR #530, opened 2017-04-19
Fix assert() so that it doesn't warn

dancrossnyc:
Rewrite this macro so it doesn't cause a spurious
warning about braces.

### DIFF
--- a/sys/src/9/port/lib.h
+++ b/sys/src/9/port/lib.h
@@ -13,7 +13,7 @@
  */
 #define nelem(x)	(sizeof(x)/sizeof((x)[0]))
 #define offsetof(s, m)	(uint64_t)(&(((s*)0)->m))
-#define assert(x)	if(x){}else _assert(#x)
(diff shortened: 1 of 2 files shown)


sys/src/9/port/lib.h
```diff
@@ -13,7 +13,7 @@
  */
 #define nelem(x)	(sizeof(x)/sizeof((x)[0]))
 #define offsetof(s, m)	(uint64_t)(&(((s*)0)->m))
-#define assert(x)	if(x){}else _assert(#x)
+#define assert(x)	do{if(!(x))_assert(#x)}while(0)
```
elbing: you missed a semicolon before closing key:
do{if(!(x)) _assert(#x);}while(0)
Also an space between if and assert, just for reading.

elbing:
Dan, rebase it.

dancrossnyc:
Done.

PR #531, opened 2017-04-19
Can set and hit a breakpoint to a stopped process

gmacd:
Once breakpoint has been hit, gdb fails because it can’t write back the instruction.  Still need to tidy up lots of this code, but at least it shows we can hit a breakpoint.  It's all very scrappy, but we're further down the path at least.

Some todos:
 - Close files
 - Modify rmem and wmem to go via map, then remove setbp call
 - Handle breakpoint note on behalf of process being debugged
 - Replace breakpoint instruction
 - Handle removal of breakpoints
etc...

### DIFF
--- a/sys/src/cmd/gdbserver/gdbstub.c
+++ b/sys/src/cmd/gdbserver/gdbstub.c
@@ -32,6 +32,8 @@
 #include <libc.h>
 #include <ureg.h>
 #include <ctype.h>
+#include <bio.h>
+#include <mach.h>
 
 #include "debug_core.h"
 #include "gdb.h"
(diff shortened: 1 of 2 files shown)


sys/src/cmd/gdbserver/gdbstub.c
```diff
@@ -134,18 +137,125 @@ p64(char *dest, uint64_t c)
...
+
+	if (syminit(textfid, &fhdr) < 0) {
+		syslog(0, "gdbserver", "syminit failed: %r");
+		return;
+	}
+
+	machbytype(fhdr.type);
+
+	// TODO close memfid
+	snprint(buf, sizeof(buf), "/proc/%d/mem", pid);
```
rminnich: or make memfid as global, open it once, use it everywhere.

gmacd: I think I'll do that - and for the ctl fid too, rather than open and close repeatedly.

PR #532, opened 2017-04-20
Really expose warnings

dancrossnyc:
But also make it so that we can continue building.
By exposing them, we see them so we can fix them.


dancrossnyc:
Correct; that's intentional, though. I want to work through these warnings and then add -Werror globally.

elbing:
Fine. Harvey is not for sissies.

PR #533, opened 2017-04-20
Kill a couple of thousand warnings.

dancrossnyc:
These are the easy ones. I've got a few hundred left.


elbing:
Directive exists for gcc and clang, but not for those purposes like in kenc. Maybe some of them could have an equivalent with _attribute_

PR #536, opened 2017-04-20
gdbserver - improvements when hitting a breakpoint

gmacd:
- wmem and rmem now use put1 and get1, so access process text and data via the map.
- dbg_remove_sw_break handles removal of active breakpoints.  When a breakpoint is hit, gdb will request that the breakpoint is removed via z0, and then reinserted when continue is requested, via Z0.  This change now behaves correctly (I think).
- Removed some unused code

The next step is to remove the breakpoint note to stop the process suiciding before we continue.

### DIFF
--- a/sys/src/cmd/gdbserver/debug_core.c
+++ b/sys/src/cmd/gdbserver/debug_core.c
@@ -69,22 +69,20 @@ int active = -1;
 char *
 arch_set_breakpoint(struct state *ks, struct bkpt *bpt)
 {
-	//char *err;
+	char *err;
 
-	//err = rmem(bpt->saved_instr, ks->threadid, bpt->bpt_addr, bpsize);
-	//if (err)
-	//	return err;
+	err = rmem(bpt->saved_instr, ks->threadid, bpt->bpt_addr, machdata->bpsize);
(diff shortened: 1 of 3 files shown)


sys/src/cmd/gdbserver/debug_core.c
```diff
@@ -69,22 +69,20 @@ int active = -1;
...
-	//	return err;
+	err = rmem(bpt->saved_instr, ks->threadid, bpt->bpt_addr, machdata->bpsize);
+	if (err)
+		return err;
 
-	setbp(bpt->bpt_addr, bpt->saved_instr);
-	//err = wmem(bpt->bpt_addr, ks->threadid, machdata->bpinst, machdata->bpsize);
-	//return err;
-	return nil;
+	err = wmem(bpt->bpt_addr, ks->threadid, machdata->bpinst, machdata->bpsize);
```
rminnich: I doubt it matters but could this call arch_remove_breakpoint?

rminnich: ah, looking at this, it can't, sorry. And it makes no sense to anyway.

PR #542, opened 2017-04-21
Expose more warnings.

dancrossnyc:
Warn on ALL the things!

Well, maybe not all, but more than before.


dancrossnyc:
That's why I turned that one back off.

PR #546, opened 2017-04-24
Use pragmas to turn off warnings about __builtin_return_address

dancrossnyc:
We need to replace this code with assembler, I think.


dancrossnyc:
I know, right? Let me make sure it boots up ok.

elbing:
boots ok for me.

dancrossnyc:
Me too, now.

PR #550, opened 2017-04-27
Fix print_func_exit() calls.

dancrossnyc:
At least one of these was out-of-order with a longjmp().
Fix the indentation on many others to make them less
obscure.

### DIFF
--- a/sys/src/cmd/aux/realemu/xec.c
+++ b/sys/src/cmd/aux/realemu/xec.c
@@ -84,8 +84,8 @@ opint(Cpu *cpu, Inst *i)
 {
 	print_func_entry();
 	cpu->trap = ar(i->a1);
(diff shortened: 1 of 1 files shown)


sys/src/cmd/aux/realemu/xec.c
```diff
@@ -84,8 +84,8 @@ opint(Cpu *cpu, Inst *i)
 {
 	print_func_entry();
 	cpu->trap = ar(i->a1);
-	longjmp(cpu->jmp, 1);
 	print_func_exit();
+	longjmp(cpu->jmp, 1);
```
rminnich: interesting. This is a coccinnelle bug.

dancrossnyc:
I wondered about that....

PR #554, opened 2017-04-27
First pass at single-step

gmacd:
- Implement single step (still needs some work)
- Removes some unused code
- Better matches the spec for qSymbol::
- Tries to fix bug when a single register value is requested

### DIFF
--- a/sys/src/cmd/gdbserver/debug_core.h
+++ b/sys/src/cmd/gdbserver/debug_core.h
@@ -15,16 +15,16 @@
 
 /* kernel debug core data structures */
 struct state {
-	unsigned long threadid;
-	int			ex_vector;
-	int			signo;
-	int			err_code;
-	int			cpu;
-	int			pass_exception;
-	char *pidname;
-	long			usethreadid;
-	void *gdbregs;
-	int gdbregsize; // determined by the amount read from /proc/pid/gdbregs
+	unsigned long	threadid;
+	int		ex_vector;
+	int		signo;
+	int		err_code;
+	int		cpu;
+	int		pass_exception;
+	char		*pidname;
+	long		usethreadid;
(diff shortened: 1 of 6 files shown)


sys/src/cmd/gdbserver/debug_core.h
```diff
@@ -15,16 +15,16 @@
 
 /* kernel debug core data structures */
 struct state {
```
rminnich: when possible it's a good idea to seperate formatting patches from real code.

sys/src/cmd/gdbserver/gdbstub.c
```diff
@@ -918,6 +934,63 @@ gdb_cmd_exception_pass(struct state *ks)
...
+		syslog(0, "gdbserver", "machdata->foll succeeded, nfoll: %d", nfoll);
+
+	for (int i = 0; i < nfoll; i++) {
+		syslog(0, "gdbserver", "Setting single-step breakpoint at: %p", foll[i]);
+		dbg_set_sw_break(ks, foll[i]);
+	}
+
+	gdb_cmd_continue(ks);
+
+	for (int i = 0; i < nfoll; i++) {
```
rminnich: don't you need to also restore the original one?

gmacd: Do you mean re-add the breakpoint be just hit?  (The one we explicitly set, not the single step bps)
If so, what happens is that once you continue, gdb single steps, then sends a packet to re-add the breakpoint, before sending another 'c'ontinue packet.  So we don't need to do it ourselved.

sys/src/cmd/gdbserver/debug_core.h
gmacd: Yeah, you're right :)

PR #556, opened 2017-04-28
Rewrite part of httpd to simplify some code.

dancrossnyc:
This change also removes a warning. It's interesting
how often warnings point out code that's not idiomatic,
or can be rewritten to make it simpler and clearer.


dancrossnyc:
While not an unappealing idea, I'm not sure I want to ping my hopes and dreams on Go at the moment....

PR #563, opened 2017-04-29
Fix the rest of the warnings.

dancrossnyc:
This fixes the rest of the warnings in the default build and
re-enable -Werror.  There may be some still lurking in the bits
of code that don't get built when we just run `util/build`
(e.g., other kernel configurations mostly) but this gets the
vast bulk of them.

Of note, RISC-V is currently broken due to unrelated missing
symbols called from the keyboard map driver.


bldybuild:
Commit 9b0368657f583b13fd6717f90156f6ecd65f56e2 is not signed-off.

Please fix these issues:

>	1. ron minnich has not signed-off 9b03686.


rminnich:
errored.

rminnich:
oh, that's just the signoff.

dancrossnyc:
No, there are legit errors. This builtin return address garbage to
implement getcallerpc(). I'm at Harvard for a conference, but will have a
look shortly.

elbing:
What if we restore wframe-address for gcc.json *only* removing pragmas until we have a proper getcallerpc?

dancrossnyc:
I've got that one fixed; the current issue is that the set of warnings
clang issues is not the same as GCC, and I don't (soon to be didn't) have
clang installed on my dev machine. That said, there's some weird junk in
Harvey; stuff like, `foo = 8192&&0xFF;`, which sets foo to `1`. I imagine
that was supposed to be `8192&0xFF`, but even that is weird; that's just
zero. plan9port sets the same thing to 8192. Ugh.

elbing:
I got that from plan9port in 2015. That would explain what you're thinking. Hey, I just thought Russ was in right... :-(

riscv/include/mach_acpi.h
```diff
@@ -0,0 +1,27 @@
+/* acpi.h */
```
rminnich: first reaction was, oh no, we won't do ACPI in riscv. Second reaction was, oh, well, thanks to linux we probably will :-(

But why are we building acpi stuff for riscv right now? I never needed to. Or was I doing it without knowing it?

sys/src/9/386/vgavesa.c
```diff
@@ -178,11 +178,12 @@ vesalinear(VGAscr *scr, int _1, int _2)
 	}
 
 	/* no pci - heuristic guess */
-	if (!havesize)
+	if(!havesize){
```
rminnich: oh wow that's nasty in the original.

sys/src/9/amd64/devacpi.c
```diff
@@ -2406,6 +2410,20 @@ static char *raw(Atable *atbl, char *start, char *end, void *unused_arg)
 	return start + len;
 }
 
+void outofyourelement(void)
```
rminnich: hey let's call this function _

?

riscv/include/mach_acpi.h
elbing: Probably when you did the squeleton following amd64 code for building it you just included it.

sys/src/9/amd64/devacpi.c
elbing: Why? Dan loves that sentence.

dancrossnyc: Not just that, but we need a symbol less-likely to collide with something else. I tried making it static and the compiler just optimizes it away and complains that the called functions aren't used; exactly what I was trying to avoid. So I had to make it extern. :-(

PR #566, opened 2017-04-29
Fix gdb writing to memory

gmacd:
Setting a variable (e.g. 'set foo=123') was broken.  This is because ebin2mem was overwriting the src rather than the dest for some reason.  wmem would then copy the dest (which hadn't been touched since the malloc...)

The first commit is the actual fix, the second tidies some things up a bit.

### DIFF
--- a/sys/src/cmd/gdbserver/gdbstub.c
+++ b/sys/src/cmd/gdbserver/gdbstub.c
@@ -518,23 +518,21 @@ hex2long(char **ptr, unsigned long *long_val)
 }
 
 /*
- * Copy the binary array pointed to by buf into mem.  Fix $, #, and
- * 0x7d escaped with 0x7d. Return -EFAULT on failure or 0 on success.
- * The input buf is overwitten with the result to write to mem.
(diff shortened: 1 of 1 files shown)


sys/src/cmd/gdbserver/gdbstub.c
```diff
@@ -518,23 +518,21 @@ hex2long(char **ptr, unsigned long *long_val)
...
- * Copy the binary array pointed to by buf into mem.  Fix $, #, and
- * 0x7d escaped with 0x7d. Return -EFAULT on failure or 0 on success.
- * The input buf is overwitten with the result to write to mem.
+ * Copy the binary array pointed to by src into dest.  Fix $, #, and
+ * 0x7d escaped with 0x7d.
  */
-static  void
-ebin2mem(unsigned char *buf, unsigned char *mem, int count)
+static void
+ebin2mem(unsigned char *src, unsigned char *dest, int count)
```
rminnich: ah what a nice idea, name things src and test so you can tell. :-)

PR #572, opened 2017-05-01
JSON: run preen on json files

rminnich:
Long overdue:
./util/preen -d=false `find . -name "*.json" -print`

Preen needs some cleanup. I never changed it since for a while util/build
was going away. But it still works.


dancrossnyc:
This is great; it's ever-so-slightly a shame that it splits up a couple of lines where it makes sense to have the compiler option as the same line as the option value, but such is life.

rminnich:
yeah, it's one of those "not ideal to a human but at least the humans don't do it by hand" deals.

PR #575, opened 2017-05-01
Improve gdb's handling of detach and kill

gmacd:
detach will start the process after removing all breakpoints.
kill will properly kill the process.

### DIFF
--- a/sys/src/cmd/gdbserver/gdbstub.c
+++ b/sys/src/cmd/gdbserver/gdbstub.c
@@ -140,9 +140,8 @@ void
 attach_to_process(int pid)
 {
 	char buf[100];
-	sprint(buf, "/proc/%d/text", pid);
 
-	// TODO close textfid?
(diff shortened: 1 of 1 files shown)


sys/src/cmd/gdbserver/gdbstub.c
```diff
@@ -718,30 +716,30 @@ gdb_cmd_binwrite(struct state *ks)
...
+	put_packet(remcom_out_buffer);
+}
+
+/* Handle the 'k' kill packet */
+static void
+gdb_cmd_kill(struct state *ks)
+{
+	sendctl(ks->threadid, "kill");
+	connected = 0;
+	print("Process %d killed by gdb\n", ks->threadid);
```
rminnich: should this be a log or fprint to 2? I don't kniow.

gmacd: I think it's useful feedback for the user, especially if the process is not visible.

PR #583, opened 2017-05-06
Apexnote

rminnich:
This fixes the apex regression test. You can try on apex if you want.

### DIFF
--- a/sys/src/regress/apexnotecont.c
+++ b/sys/src/regress/apexnotecont.c
@@ -28,12 +28,19 @@ _notetramp(int sig, void (*hdlr)(int, char*, struct Ureg*), struct Ureg *u)
 	if(nstack >= MAXSIGSTACK)
 		noted(1);	/* nesting too deep; just do system default */
 	p = &pcstack[nstack];
+print("a\n");
 	p->restorepc = u->ip;
+print("b\n");
 	p->sig = sig;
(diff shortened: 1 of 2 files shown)


sys/src/regress/apexnotecont.c
```diff
@@ -28,12 +28,19 @@ _notetramp(int sig, void (*hdlr)(int, char*, struct Ureg*), struct Ureg *u)
...
 	p->restorepc = u->ip;
+print("b\n");
 	p->sig = sig;
+print("c\n");
 	p->hdlr = hdlr;
+print("d\n");
 	p->u = u;
+print("e\n");
 	nstack++;
+print("f\n");
```
dancrossnyc: Are all of these print()'s still necessary? This feels like debugging code that leaked into a commit.

rminnich: It's a regression test so I think we should leave them in.

dancrossnyc: Okay; that makes sense.

rminnich: I don't know I might be full of it ... is it ok in a test situation to allow arbitrary text on stdout? It's super handy in this case for isolating where the failure is but ... your ideas here are most welcome. Put it on stdout, stderr, or ...

PR #585, opened 2017-05-10
Implement get_reg for gdbserver

gmacd:
Rather important for getting the stack pointer to step out of functions.
Also tidy the formatting of some error messages that showed up alongside the bug.

Seems the list of registers is duplicated in sys/src/9/amd64/dat.h
We should probably pick one place for this to avoid the duplication.

### DIFF
--- a/sys/src/cmd/gdbserver/gdbstub.c
+++ b/sys/src/cmd/gdbserver/gdbstub.c
@@ -1,4 +1,4 @@
-/*
+	/*
  * Kernel Debug Core
  *
  * Maintainer: Jason Wessel <jason.wessel@windriver.com>
@@ -55,6 +55,8 @@ int notefid;
 int debug = 0;
 int attached_to_existing_pid = 0;
 Map* cormap;
+
+static struct state ks;
 
 /* support crap */
 /*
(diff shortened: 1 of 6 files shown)


sys/src/cmd/gdbserver/gdbstub.c
```diff
@@ -56,13 +56,43 @@ int debug = 0;
...
+static struct state ks;
+
 /* support crap */
 /*
  * GDB remote protocol parser:
  */
 
 static char *hex_asc = "0123456789abcdef";
 
+char* regstrs[] = {
```
rminnich: at some point we'll want this definition to go into a file named amd64.c, and reference it in the .json as $ARCH, but that can come later.

sys/src/cmd/gdbserver/gdbstub.c
```diff
@@ -604,14 +634,23 @@ int_to_threadref(unsigned char *id, int value)
 uint64_t
 get_reg(Map *map, char *reg)
 {
-	//uint64_t v;
-	//int ret = get8(map, reg, &v);
+	(void)map;
+
+	int reg_idx = -1;
+	for (int i = 0; i < DBG_MAX_REG_NUM; i++) {
```
rminnich: dbg_max_reg_num would be a constant defined in, hmm, ureg.h?

sys/src/cmd/gdbserver/regsamd64.c
```diff
@@ -104,6 +104,21 @@ gdb_cmd_reg_set(struct state *ks)
 }
 
 uint64_t
```
rminnich: ooh nice the way you broke this out.

sys/src/cmd/gdbserver/gdbstub.c
gmacd: Agreed.

PR #588, opened 2017-05-15
Add a script that will list all UFS files that break spatch

gmacd:
Gives us the following list of:
../ufs/dirhash.h
../ufs/extattr.h
../ufs/inode.h
../ufs/quota.h
../ufs/ufs_dirhash.c
../ufs/ufs_quota.c
../ffs/ffs_extern.h
../ffs/ffs_snapshot.c
../ffs/ffs_softdep.c
../ffs/ffs_vfsops.c
../ffs/softdep.h

The script can be adapted for future ports.

### DIFF
--- a/sys/src/libufs/port-scripts/list-spatch-breakers.sh
+++ b/sys/src/libufs/port-scripts/list-spatch-breakers.sh
@@ -3,7 +3,7 @@
 # Prints out a list of all UFS files that cannot be fully parsed by spatch
 
 for file in ../{ufs,ffs}/*.[ch]; do
-	if [[ $(spatch --parse-c $file 2>&1 | grep ^bad:) ]]; then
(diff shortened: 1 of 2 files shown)


rminnich:
I have to give this a try.

rminnich:
The script fails for me. What shell is it?
sh bla bla gives me
[[: not found

gmacd:
Ah, must be a bash thing.  Do we mind if we use bash, or only sh?

I can change either way.

rminnich:
if it's a bash thing then the #! needs to be /bin/bash, not /bin/sh

elbing:
The problem seems that sh is a symlink to dash in many linuxes and a bourne shell in BSDs. We should do shell independent or check env:
#!/usr/bin/env bash

elbing:
Also, when harvey was built through a bash script in early days, "for" loops were warned by Ron as a future generator of troubles. One more time, as usually, he's in right (man, the first time you were wrong I'll do a party).

PR #590, opened 2017-05-17
hash: add hash function from aki

rminnich:
this includes a test. It needs a fix in the formatting: what is %z%u in the
plan 9 world?


elbing:
```c
size_t x = ...;
ssize_t y = ...;
printf("%zu\n", x);  // prints as unsigned decimal
printf("%zx\n", x);  // prints as hex
printf("%zd\n", y);  // prints as signed decimal
```
It's a C99 addition, but really it can be removed.

PR #604, opened 2017-05-20
Trying to get ufs_bmap.c to build (part 1)

gmacd:
Pull some defines and typedefs from FreeBSD into freebsd_util.h.
Remove some defines that wouldn't do anything in C.
Fix includes and signatures in ufs_bmap.c.

### DIFF
--- a/sys/src/libufs/ufs/freebsd_util.h
+++ b/sys/src/libufs/ufs/freebsd_util.h
@@ -1,0 +1,60 @@
+/*-
+ * Copyright (c) 1991, 1993
+ *	The Regents of the University of California.  All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. Neither the name of the University nor the names of its contributors
+ *    may be used to endorse or promote products derived from this software
+ *    without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE REGENTS AND CONTRIBUTORS ``AS IS'' AND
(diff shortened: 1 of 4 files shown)


sys/src/libufs/ufs/freebsd_util.h
```diff
@@ -0,0 +1,60 @@
...
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ *
+ *	@(#)queue.h	8.5 (Berkeley) 8/20/94
+ * $FreeBSD$
+ */
+
+#define	TAILQ_HEAD(name, type)						\
```
rminnich: These are kinda nasty and definitely not plan 9 style. Is there some way to adapt the code to not use these? Or change them to look like plan 9 code?

gmacd: It depends on our approach - do we try to get it building first, with as few changes as possible, or make more drastic changes as we go along?
I'd initially been aiming to just get it building, but came across bigger problems when I reached the 'buf' struct, which is a monster that looks like it would pull in even more FreeBSD code.

rminnich: How many places in the code is this struct used? If it is used once, then this macro is not that useful. It might be worth looking at other bits of code to get some idea of how much the macro is needed vs. just changing the structs. 
bus, vnode, vnops, will all be pretty hard.

gmacd: 12 references to TAILQ_HEAD, 23 to LIST_HEAD.

gmacd: I agree we should replace them out eventually, but it seems we have bigger fish to fry.

elbing:
What about if we get in first and clean it after (a la acpi or freetype)? Having anything building and working give us a permanent test for evey cleaning change: it works or not.

rminnich:
I agree. Also, since Graham is doing real work, I should not be getting so picky.

PR #606, opened 2017-05-23
Adding more programs from src/cmd

elbing:
Added:
execnet
bc (doesn't work properly, help needed)
units
vt (segattach always returns (void *)-1)

Removed and added to APEX repo:
bzip2recover
compress
eqn


sys/src/cmd/bc/bc.y
```diff
@@ -29,7 +29,7 @@
 	char*	pre;
 	char*	post;
 
-	long	peekc = -1;
+	int32_t	peekc = -1;
```
rminnich: This really ought to be an int if possible. Do we really have to say int32 in this case? It's just a char.

sys/src/cmd/bc/bc.y
```diff
@@ -77,11 +77,12 @@
 	void	pp(char*);
 	void	routput(char*);
 	void	tp(char*);
+	int yylex(void);
 	void	yyerror(char*, ...);
 	int	yyparse(void);
 
 	typedef	void*	pointer;
-	#pragma	varargck	type	"lx"	pointer
+	/* #pragma	varargck	type	"lx"	pointer */
```
rminnich: If this will never ever be used, kill it.

sys/src/cmd/bc/bc.y
```diff
@@ -773,7 +773,7 @@ cpeek(int c, int yes, int no)
 int
 getch(void)
 {
-	long ch;
+	int32_t ch;
```
rminnich: getch returns an int. Why is this int32_t?

sys/src/cmd/units/units.y
```diff
@@ -18,7 +18,7 @@ typedef	struct	Prefix	Prefix;
 struct	Node
 {
 	double	val;
-	schar	dim[Ndim];
+	uint8_t	dim[Ndim];
```
rminnich: schar is signed char. Why is this unsigned?

sys/src/cmd/bc/bc.y
elbing: How could a char be a long in the original?

sys/src/cmd/bc/bc.y
elbing: Getting a knife...

sys/src/cmd/bc/bc.y
elbing: In harvey we converted long to int32 and vlong to int64. I'll put an int there.

sys/src/cmd/units/units.y
elbing: Because my brain saw there  "short char". I never saw "schar" until now (2017).

elbing:
Any idea about why bc doesn't work? I doubt that types could mark a difference.

bldybuild:
Commit cd814b7d8264f9fcf4383be064bf4d84d5389046 is not signed-off.

Please fix these issues:

>	1. Álvaro Jurado has not signed-off cd814b7.

PR #611, opened 2017-06-08
Add dummy mount file and directory to devufs

gmacd:
Also noticed that the strlen declaration was missing const modifier.  Others are missing, but can be added separately.


sys/src/9/port/lib.h
```diff
@@ -40,7 +40,7 @@ extern	char*	strlcpy(char*, char*, int32_t);
 extern	char*	strncpy(char*, char*, int32_t);
 extern	int	strncmp(char*, char*, int32_t);
 extern	char*	strrchr(char*, int);
-extern	int32_t	strlen(char*);
+extern	int32_t	strlen(const char*);
```
rminnich: consider making this an int. Everything else is.

gmacd: Ok, I'm going to have a pass over the string functions next, so we at least have matching signatures (and make this an int).

PR #616, opened 2017-06-11
ufs - first attempt to actually mount file in devufs

gmacd:

### DIFF
--- a/sys/src/9/port/devufs.c
+++ b/sys/src/9/port/devufs.c
@@ -139,22 +139,41 @@ ufsread(Chan *c, void *a, int32_t n, int64_t offset)
 }
 
 static void
-mount(const char* path)
+mount(char* a, int32_t n)
 {
-	if (!path || strlen(path) == 0) {
-		error(Enonexist);
+	Proc *up = externup();
+
+	Cmdbuf* cb = parsecmd(a, n);
+	if (waserror()) {
+		free(cb);
+		nexterror();
+		print("couldn't parse mount path: %s: %r", a);
+		return;
+	}
+	poperror();
+
+	// TODO Make namec and related functions accept const char*
+	Chan* c = namec(cb->buf, Aopen, ORDWR, 0);
+	free(cb);
+	cb = nil;
+	if (waserror()) {
+		cclose(c);
+		nexterror();
+		print("couldn't open file %s to mount: %r", cb->buf);
+		return;
+	}
+	poperror();
+
+	struct mount* mp = newufsmount(c);
+	int rcode = ffs_mount(mp);
+	if (rcode != 0) {
+		// TODO translate error to string
+		print("failed to mount %s with error code: %d", cb->buf, rcode);
+		releaseufsmount(mp);
 		return;
 	}
 
-	print("Mount %s\n", path);
-	struct mount* mp = mallocz(sizeof(struct mount), 1);
-	int rcode = ffs_mount(mp);
-	if (rcode != 0) {
-		// TODO translate error
-		print("Error mounting %s\n", path);
-	} else {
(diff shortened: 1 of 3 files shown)


sys/src/9/port/devufs.c
```diff
@@ -139,22 +139,41 @@ ufsread(Chan *c, void *a, int32_t n, int64_t offset)
...
+
+	// TODO Make namec and related functions accept const char*
+	Chan* c = namec(cb->buf, Aopen, ORDWR, 0);
+	free(cb);
+	cb = nil;
+	if (waserror()) {
+		cclose(c);
+		nexterror();
+		print("couldn't open file %s to mount: %r", cb->buf);
 		return;
```
rminnich: you don't need the return, nexterror() will not return. This also means your print won't be seen.

sys/src/9/port/devufs.c
```diff
@@ -139,22 +139,41 @@ ufsread(Chan *c, void *a, int32_t n, int64_t offset)
...
 	}
+	poperror();
 
-	print("Mount %s\n", path);
-	struct mount* mp = mallocz(sizeof(struct mount), 1);
+	struct mount* mp = newufsmount(c);
```
rminnich: Is this a storage leak?

sys/src/9/port/devufs.c
```diff
@@ -139,22 +139,41 @@ ufsread(Chan *c, void *a, int32_t n, int64_t offset)
...
 	int rcode = ffs_mount(mp);
```
rminnich: could you have ffs_mount call error? this would really simplify the logic here. This is kind of a mix of error()-style and Unix-style logic. It would be nice if you could use error() in the plan 9 style. This would mean changing ffs_mount to call error()

sys/src/9/port/devufs.c
```diff
@@ -167,7 +186,7 @@ ufswrite(Chan *c, void *a, int32_t n, int64_t offset)
 
 	switch (QID(c->qid)) {
 	case Qmount:
-		mount((char*)a);
+		mount((char*)a, n);
```
rminnich: The general way to do this is set n to desired value (or don't set n) and then break; and fall out of the switch and exit.

sys/src/libufs/ufs_harvey.c
```diff
@@ -0,0 +1,46 @@
...
+
+#include <u.h>
+#include <libc.h>
+
+#include "ufs_harvey.h"
+#include "libufs.h"
+
+
+struct mount*
+newufsmount(Chan* c)
```
rminnich: can this fail? the code above doesn't handle failure.

sys/src/libufs/ufs_harvey.c
```diff
@@ -0,0 +1,46 @@
...
+{
+	struct vnode* vn = mallocz(sizeof(struct vnode), 1);
+	return vn;
+}
+
+void
+releaseufsmount(struct mount* mp)
+{
+	// TODO HARVEY - This assumes no sharing
+	free(mp);
```
rminnich: you should probably put a Ref in struct mount, and get that handling right now. This is going to be painful to debug if you don't get that right from the start.

sys/src/libufs/ufs_harvey.c
```diff
@@ -0,0 +1,46 @@
...
+}
+
+void
+releaseufsvnode(struct vnode* vn)
+{
+	// TODO HARVEY - This assumes no sharing
```
rminnich: same comment. Consider a Ref.

gmacd:
Thanks, I'll take a look. I screwed up the pr, and missed another commit I meant to cherry pick.

PR #619, opened 2017-06-15
Remove dead code in newfs

driusan:
Remove code that was guarded by an #if 0 block in newfs, and references to
an undefined __FBSDID macro.

Signed-off-by: Dave MacFarlane <driusan@gmail.com>


elbing:
Please, rebase it too, we need a policy about the order of merging.

driusan:
done

PR #623, opened 2017-06-18
newfs: Remove FreeBSD headers and add header from libufs

driusan:
- Add header from libufs which defines some FreeBSD types for Harvey rather
than duplicating the effort, and '#if 0'ed out the FreeBSD headers which
don't exist.
- Add a couple FreeBSD ufs types needed for newfs to libufs header

Signed-off-by: Dave MacFarlane <driusan@gmail.com>

### DIFF
--- a/sys/src/cmd/newfs/newfs.c
+++ b/sys/src/cmd/newfs/newfs.c
@@ -36,6 +36,17 @@
  * SUCH DAMAGE.
  */
 
+#include <u.h>
+#include <libc.h>
+
+
+#include <ufs/dir.h>
+#include <ufs/dinode.h>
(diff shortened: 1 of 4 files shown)


sys/src/cmd/newfs/newfs.c
```diff
@@ -36,6 +36,11 @@
  * SUCH DAMAGE.
  */
 
+#include <u.h>
+#include <libc.h>
+
+#include <ufs/freebsd_util.h>
+#if 0
```
rminnich: Later we should just remove these.

driusan: I agree, but for now they're handy to have for reference

PR #624, opened 2017-06-19
devufs can now read the superblock when mounting

gmacd:

### DIFF
--- a/sys/include/libufs.h
+++ b/sys/include/libufs.h
@@ -9,26 +9,29 @@
  */
 
 
-struct Chan;
-struct vnode;
-struct thread;
+typedef struct Chan Chan;
+typedef struct ufsmount ufsmount;
+typedef struct vnode vnode;
+typedef struct thread thread;
+
 
 /* Wrapper for a UFS mount.  Should support reading from both kernel and user
  * space (eventually)
  */
-struct mount {
-	struct ufsmount*	mnt_data;
-	struct Chan*		chan;
-	int32_t			(*read)(struct mount*, void*, int32_t, int64_t);
-};
+typedef struct MountPoint {
+	ufsmount	*mnt_data;
+	Chan		*chan;
+	int32_t		(*read)(struct MountPoint*, void*, int32_t, int64_t);
+} MountPoint;
 
-struct mount* newufsmount(
-	struct Chan* c,
-	int32_t (*read)(struct mount*, void*, int32_t, int64_t));
 
-struct vnode* newufsvnode();
+MountPoint *newufsmount(
+	Chan *c,
+	int32_t (*read)(MountPoint*, void*, int32_t, int64_t));
 
-void releaseufsmount(struct mount* mp);
-void releaseufsvnode(struct vnode* vn);
+vnode* newufsvnode();
 
-int ffs_mount(struct mount *mp);
+void releaseufsmount(MountPoint *mp);
+void releaseufsvnode(vnode *vn);
+
+int ffs_mount(MountPoint *mp);
--- a/sys/include/libufs.h
+++ b/sys/include/libufs.h
@@ -10,11 +10,22 @@
 
 
 struct Chan;
-struct mount;
 struct vnode;
+struct thread;
 
+/* Wrapper for a UFS mount.  Should support reading from both kernel and user
+ * space (eventually)
+ */
+struct mount {
+	struct ufsmount*	mnt_data;
+	struct Chan*		chan;
+	int32_t			(*read)(struct mount*, void*, int32_t, int64_t);
+};
 
-struct mount* newufsmount(struct Chan* c);
+struct mount* newufsmount(
+	struct Chan* c,
+	int32_t (*read)(struct mount*, void*, int32_t, int64_t));
+
 struct vnode* newufsvnode();
 
 void releaseufsmount(struct mount* mp);
--- a/sys/src/9/port/devufs.c
+++ b/sys/src/9/port/devufs.c
@@ -138,10 +138,16 @@ ufsread(Chan *c, void *a, int32_t n, int64_t offset)
 	return n;
 }
 
+static int32_t
+chanread(struct mount* mp, void *p, int32_t n, int64_t off)
+{
+	return mp->chan->dev->read(mp->chan, p, n, off);
+}
+
 static struct mount*
 mountufs(Chan* c)
 {
-	struct mount* mp = newufsmount(c);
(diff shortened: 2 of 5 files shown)


sys/include/libufs.h
```diff
@@ -10,11 +10,22 @@
...
 struct Chan;
-struct mount;
 struct vnode;
+struct thread;
 
+/* Wrapper for a UFS mount.  Should support reading from both kernel and user
+ * space (eventually)
+ */
+struct mount {
+	struct ufsmount*	mnt_data;
```
rminnich: we need to start making this more like Plan 9 code.
typedef Struct ufsmount *Ufsmount
then declare it as Ufsmount *mnt_data;

sys/include/libufs.h
```diff
@@ -10,11 +10,22 @@
...
+	struct Chan*		chan;
```
rminnich: Chan *chan;

sys/include/libufs.h
```diff
@@ -10,11 +10,22 @@
...
+	int32_t			(*read)(struct mount*, void*, int32_t, int64_t);
```
rminnich: what's the reason for this? You have the chan. Why is that not sufficient?

sys/include/libufs.h
```diff
@@ -10,11 +10,22 @@
...
+};
+
+struct mount* newufsmount(
+	struct Chan* c,
```
rminnich: Chan *c, not struct.

sys/include/libufs.h
```diff
@@ -10,11 +10,22 @@
...
+	int32_t (*read)(struct mount*, void*, int32_t, int64_t));
 
-struct mount* newufsmount(struct Chan* c);
 struct vnode* newufsvnode();
```
rminnich: You need to create
typedef struct vnode Vnode.

Or better struct Vnode.
typedef struct Vnode Vnode;

sys/include/libufs.h
```diff
@@ -10,11 +10,22 @@
...
 
 void releaseufsmount(struct mount* mp);
```
rminnich: Mount *mp

sys/src/9/port/devufs.c
```diff
@@ -138,10 +138,16 @@ ufsread(Chan *c, void *a, int32_t n, int64_t offset)
 	return n;
 }
 
+static int32_t
+chanread(struct mount* mp, void *p, int32_t n, int64_t off)
```
rminnich: I don't get why you need this. Why isn't dev->read enough? 

FYI the Plan 9 way to do this is you store the type in the QID and case out on that in dev->read.

sys/src/9/port/devufs.c
```diff
@@ -138,10 +138,16 @@ ufsread(Chan *c, void *a, int32_t n, int64_t offset)
...
+{
+	return mp->chan->dev->read(mp->chan, p, n, off);
+}
+
 static struct mount*
 mountufs(Chan* c)
 {
-	struct mount* mp = newufsmount(c);
+	struct mount* mp = newufsmount(c, chanread);
```
rminnich: Mount * but anyway we need to revisit all this chanread stuff.

sys/src/libufs/ffs/ffs_vfsops.c
```diff
@@ -38,17 +38,19 @@
...
 //#include "../ffs/ffs_extern.h"
 
 #include "libufs.h"
+#include "ufs_harvey.h"
 
-#if 0
-static uma_zone_t uma_inode, uma_ufs1, uma_ufs2;
+//static uma_zone_t uma_inode, uma_ufs1, uma_ufs2;
 
 static int	ffs_mountfs(struct vnode *, struct mount *, struct thread *);
```
rminnich: Vnode *, Mount *, Thread *

This is not linux :-)

sys/src/libufs/ffs/ffs_vfsops.c
```diff
@@ -85,9 +87,10 @@ static const char *ffs_opts[] = { "acls", "async", "noatime", "noclusterr",
 int
 ffs_mount(struct mount *mp)
 {
+	struct vnode *devvp = nil;
```
rminnich: Moutn *
Vnode *

well you get my drift by now I bet.

sys/src/libufs/ufs_harvey.c
```diff
@@ -17,11 +17,14 @@
 
 
 struct mount*
-newufsmount(Chan* c)
+newufsmount(
+	struct Chan* c,
```
rminnich: Chan *c

sys/include/libufs.h
gmacd: I'm trying to make the library usable for kernel or user code.  I could only see a way to get access to the channel's internal dev member (in portdat.h) in the kernel code.  Also, I assumed people would usually open a file outside the kernel using 'open'.  This way permits either, assuming we eventually make the Chan a union with a fid (or do something else).

sys/include/libufs.h
gmacd: Ok - I'll look at a spatch to rename ufsmount in a later PR and make the typedef change for this one.

sys/include/libufs.h
gmacd: I'll make a spatch PR for this soon.

sys/src/libufs/ffs/ffs_vfsops.c
gmacd: Hmm, one more time?

sys/include/libufs.h
rminnich: Were it me, I would make it work for one thing, and worry about the other later when we've figured out how it all works. The reason is that this is getting kinda nasty at this point, doing things you don't want to do in kernel for the eventual and maybe never happens user mode case. My belief is we should make this right for kernel first, and then figure out the user case if and when we need it.

gmacd: Alright - let me look into it.

elbing:
looks fine for me now.

sys/include/libufs.h
```diff
@@ -9,15 +9,29 @@
  */
 
 
-struct Chan;
-struct mount;
-struct vnode;
+typedef struct Chan Chan;
```
rminnich: it's weird that you need this but ...

gmacd:
I haven't finished yet - still working on removing the read field.

PR #626, opened 2017-06-22
Read superblock when mounting UFS2 file system

gmacd:
Now focusing only on reading from within kernel.  If it ever needs to be done from user-space, we need to unpick it.


sys/src/9/ufs/ufs_harvey.h
```diff
@@ -9,19 +9,52 @@
...
-	//struct mount*	v_mount;
-};
+typedef struct vnode {
+	inode	*v_data;
+	//MountPoint	*v_mount;
+} vnode;
+
+
+// Not sure we even need this - if not we can remove it later.
+typedef struct thread {
```
rminnich: it doesn't make a lot of sense in the kernel ... in harvey, there's no such thing as a kernel level thread. Just processes.

sys/src/9/ufs/ffs/ffs_vfsops.c
```diff
@@ -773,13 +790,16 @@ ffs_mountfs (struct vnode *devvp, struct mount *mp, struct thread *td)
...
 			break;
-		brelse(bp);
-		bp = nil;
 	}
+
 	if (sblock_try[i] == -1) {
-		error = EINVAL;		/* XXX needs translation */
+		//error = EINVAL;		/* XXX needs translation */
+		// TODO HARVEY error string?
+		error = -1;
```
rminnich: should be using error() here but we can always fix that up later. The goto stuff is not really how it should be done but I realize this comes from unix.

sys/src/9/ufs/ufs_harvey.h
gmacd: Yeah, I was wondering if we'd need to pass some different info down here - e.g. a process ID, so I left it in for now.

sys/src/9/ufs/ffs/ffs_vfsops.c
gmacd: Agreed, though in these cases I'm reluctant to change too much right now.  Also, I don't want to break the flow of code for error handling, especially for deeper errors that might bubble up.

PR #632, opened 2017-06-26
UFS: More mount code

gmacd:


sys/src/9/ufs/ffs/ffs_vfsops.c
```diff
@@ -905,28 +905,28 @@ ffs_mountfs (vnode *devvp, MountPoint *mp, thread *td)
...
 	    (nmp = vfs_getvfs(&mp->mnt_stat.f_fsid))) {
 		if (nmp)
 			vfs_rel(nmp);
 		vfs_getnewfsid(mp);
-	}
-	mp->mnt_maxsymlinklen = fs->fs_maxsymlinklen;
-	MNT_ILOCK(mp);
+	}*/
+	//mp->mnt_maxsymlinklen = fs->fs_maxsymlinklen;
+	qlock(&mp->mnt_lock);
```
elbing: What does MNT_ILOCK? I mean if it does some kind of ilock, I'm not sure if having some kind of interrupt qlock would work fine. If you tested and worked, just asking, anyway.

gmacd: MNT_ILOCK is a wrapper to lock the mutex on the mountpoint.  I take FreeBSDs mutex to be equivalent to our qlock, since neither are spinlocks.  Not sure how they'd behave with interrupts tbh.

gmacd: Add I understand now, a spinlock (Lock) shouldn't be used in an interrupt in case of a nested interrupt. Not aware of any problems with QLock.

PR #635, opened 2017-07-09
UFS Add ctl file for devufs mountpoints

gmacd:
echo unmount > /dev/ufs/0/ctl
will now unmount mountpoint 0.

Also embedding the mountpoint ID in the qid path, as seen in devproc.

### DIFF
--- a/sys/src/9/port/devufs.c
+++ b/sys/src/9/port/devufs.c
@@ -19,25 +19,40 @@
 
 
 enum {
-	Qdir = 0,		// #U
-	Qufsdir,		// #U/ufs
-	Qmount,			// #U/ufs/mount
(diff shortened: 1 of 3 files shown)


sys/src/9/port/devufs.c
```diff
@@ -196,21 +247,56 @@ mount(char* a, int32_t n)
...
+{
+	Proc *up = externup();
+
+	if (mountpoint == nil) {
+		error(Eunmount);
+	}
+
+	Cmdbuf* cb = parsecmd(a, n);
+	if (waserror()) {
+		print("couldn't parse command: %s: %r", a);
```
elbing: you can use error("string"), see other waserror in port dir.

gmacd: Yeah, I think I need to use more specific error messages.

PR #636, opened 2017-07-11
Start to expose ufs_lookup, and some tidying of UFS code

gmacd:


sys/src/9/ufs/ffs/ffs_vfsops.c
```diff
@@ -35,11 +35,11 @@
 #include "dat.h"
 #include "../../port/portfns.h"
 
-#define _KERNEL
-
 #include "freebsd_util.h"
 #include "ufs_harvey.h"
 
+#define _KERNEL
```
rminnich: The sooner we can get rid of this the better :-)

gmacd: Agreed, might be worth getting rid of some of them in the next commit.

PR #639, opened 2017-07-30
Get root UFS inode, add ability to test from devufs, UFS2 only

gmacd:
Quite a few changes here.  My current goal is to be able to resolve a path down to an inode by enabling the lookup code.  The first step is to get the root inode though, so that's what's happening here.

- Read the root inode.
- Add simple test command (```echo test > /dev/ufs/0/ctl```) to run some tests on mountpoint 0.  Currently just reads in the root inode and reports if there's an error.
- Decided to remove any UFS1-only code, so made a start in some places.  Rather than go through the vops interface, we'll try to call the UFS functions directly.
- Still need to implement locking.  This requires a bit of thought, because there's probably not a direct mapping between the FreeBSD locking mechanisms used here, and those in Harvey.
- There are various TODOs for using a pool of vnodes/inodes, hashing, better use of buffers for reading, etc.
- I had assumed that inode was a direct representation of what's on disk, but it turns out I was wrong - the inode is read in from disk, then the inode struct is built based on some of the data.  We should be able to get rid of the vnode layer at some point and work directly on inodes.  This will mean managing pools of inodes, moving locks onto that structure, etc.
- MountPoint needs to keep track of vnode/inodes.


sys/src/9/ufs/ufs/ufs_inode.c
```diff
@@ -47,7 +54,19 @@
...
 #include <ufs/ufs/gjournal.h>
-#endif
+#endif*/
+
+
+/*
+ * Conversion tables for conversion from vnode types to inode formats
+ * and back.
+ */
+static Vtype iftovt_tab[16] = {
```
rminnich: wow, they have all these types? The only ones in there should be DIR, REG and, maybe, LNK

sys/src/9/ufs/ufs/ufs_vnops.c
```diff
@@ -2394,28 +2407,31 @@ ufs_pathconf (struct vop_pathconf_args *ap)
...
 
 	vp = *vpp;
 	ip = VTOI(vp);
-	vp->v_type = IFTOVT(ip->i_mode);
-	if (vp->v_type == VFIFO)
-		vp->v_op = fifoops;
-	ASSERT_VOP_LOCKED(vp, "ufs_vinit");
+	vp->v_type = ifmt_to_vtype(ip->i_mode);
+	// TODO HARVEY Locking
+	//ASSERT_VOP_LOCKED(vp, "ufs_vinit");
```
rminnich: should almost certainly be just a Qlock.

sys/src/9/ufs/ufs_harvey.c
```diff
@@ -48,3 +48,85 @@ releaseufsvnode(vnode *vn)
...
+{
+	vnode *vp = nil;
+	//struct lock_object *lo;
+	//static int cyclecount;
+	//int error = 0;
+
+//alloc:
+	vp = (vnode *)smalloc(sizeof(vnode));
+
+	// TODO HARVEY Revisit locking of vnodes
```
rminnich: note: get the locking right from the start, your life is going to be pretty awful otherwise.

sys/src/9/ufs/ufs_harvey.h
```diff
@@ -79,6 +69,60 @@ typedef struct ComponentName {
...
+ */
+//#define	VI_MOUNT	0x0020	/* Mount in progress */
+//#define	VI_DOOMED	0x0080	/* This vnode is being recycled */
+//#define	VI_FREE		0x0100	/* This vnode is on the freelist */
+//#define	VI_ACTIVE	0x0200	/* This vnode is on the active list */
+//#define	VI_DOINGINACT	0x0800	/* VOP_INACTIVE is in progress */
+//#define	VI_OWEINACT	0x1000	/* Need to call inactive */
+
+#define	VV_ROOT		0x0001	/* root of its filesystem */
+//#define	VV_ISTTY	0x0002	/* vnode represents a tty */
```
rminnich: this one must certainly die :-) I'm amazed that 'vnode represents a tty' is even a thing ...

sys/src/9/ufs/ufs/ufs_inode.c
gmacd: #define	IFMT		0170000		/* Mask of file type. */
#define	IFIFO		0010000		/* Named pipe (fifo). */
#define	IFCHR		0020000		/* Character device. */
#define	IFDIR		0040000		/* Directory file. */
#define	IFBLK		0060000		/* Block device. */
#define	IFREG		0100000		/* Regular file. */
#define	IFLNK		0120000		/* Symbolic link. */
#define	IFSOCK		0140000		/* UNIX domain socket. */
#define	IFWHT		0160000		/* Whiteout. */

They have a few.  Hopefully we can trim it back a bit without losing UFS2 compatibility.

sys/src/9/ufs/ufs_harvey.c
gmacd: Totally agree.

sys/src/9/ufs/ufs_harvey.h
gmacd: Hopefully we'll get to the point where we can just delete a lot of this stuff :)

sys/src/9/ufs/ufs/ufs_vnops.c
gmacd: It needs to be a lock that allows recursion.  For some reason I though qlocks don't, but looking at the code, it seems they'd be fine.

gmacd: ...although the lock sometimes changes from shared to exclusive depending on the context.

gmacd: ...although disabling sharing only seems to happen in the quota code, which we don't really care about.

gmacd: So maybe qlock is fine.

PR #642, opened 2017-08-11
UFS Mountpoint vnode management, sharing

gmacd:
- ffs_vgetf will get an existing vnode or pick one from freelist
- vnodes for same inode are now shared and reference counted
- freelist of vnodes maintained on mountpoint
- vnode lists on mountpoint protected by qlock
- added stats file to mountpoint
- started to tidy up code with ufs_ext.h for externally facing definitions

### DIFF
--- a/sys/src/9/ufs/ufs_harvey.h
+++ b/sys/src/9/ufs/ufs_harvey.h
@@ -10,32 +10,11 @@
 
 
 typedef struct Chan Chan;
+typedef struct inode inode;
+typedef struct MountPoint MountPoint;
+typedef struct thread thread;
 typedef struct ufsmount ufsmount;
 typedef struct vnode vnode;
-typedef struct thread thread;
-typedef struct inode inode;
-
-
-/*
- * filesystem statistics
- */
-typedef struct statfs {
-	uint64_t f_iosize;		/* optimal transfer block size */
-} StatFs;
-
-/* Wrapper for a UFS mount.  Should support reading from both kernel and user
- * space (eventually)
- */
-typedef struct MountPoint {
-	ufsmount	*mnt_data;
-	Chan		*chan;
-	int		id;
-	StatFs		mnt_stat;		/* cache of filesystem stats */
-	int		mnt_maxsymlinklen;	/* max size of short symlink */
-
-	uint64_t	mnt_flag;		/* (i) flags shared with user */
-	QLock		mnt_lock;		/* (mnt_mtx) structure lock */
-} MountPoint;
 
 
 // Not sure we even need this - if not we can remove it later.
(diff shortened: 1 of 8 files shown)


sys/src/9/ufs/ufs_harvey.h
```diff
@@ -120,16 +99,21 @@ typedef enum vtype Vtype;
...
+	Ref		ref;		/* Refcount */
+	QLock		v_vnlock;	/* u pointer to vnode lock */
+
 	/*
 	 * Fields which define the identity of the vnode.  These fields are
 	 * owned by the filesystem (XXX: and vgone() ?)
 	 */
-	const char 	*v_tag;			/* u type of underlying data */
 	inode		*v_data;
 	MountPoint	*v_mount;
```
rminnich: general comment here and elsewhere. 

FWIW the plan 9 style tended in later years to CamelCase, not camel_case. I.e. more like go, less like traditional Unix. Also, for example, the v->v_mount struct element thing was gradually discarded in favor of just the name, i.e. v_mount would become just Mount or mount. The v_ type stuff really goes back to the time when any struct pointer could deref any struct type (true! see dmr's early paper in the v6 docs). At some point I'd favor dropping all the v_ etc. stuff in these structs but that's your call of course.

sys/src/9/ufs/ufs_mountpoint.c
```diff
@@ -0,0 +1,184 @@
...
+        return mp;
+}
+
+void
+releaseufsmount(MountPoint *mp)
+{
+        // No need to unlock later, since we're freeing mp
+        qlock(&mp->vnodes_lock);
+
+        // TODO HARVEY What if there are referenced vnodes?
```
rminnich: you'd probably have a ref in the mountpoint and sleep on it until it went to zero maybe? Not sure.

sys/src/9/ufs/ufs_mountpoint.c
```diff
@@ -0,0 +1,184 @@
...
+
+vnode*
+getfreevnode(MountPoint *mp)
+{
+        qlock(&mp->vnodes_lock);
+
+        if (mp->free_vnodes == nil) {
+                mp->free_vnodes = alloc_freelist();
+                if (mp->free_vnodes == nil) {
+                        return nil;
```
rminnich: does this leave the qlock locked?

sys/src/9/ufs/ufs_harvey.h
gmacd: Would that be CamelCase starting with a capital for struct fields?  I've seen some examples of fields starting with a capital letter, but wasn't sure if it was an aberration...  I'd usually prefer starting with lower case (camelCase) but always prefer to be consistent at least.

The v_* fields were copied directly, but I'm feeling more relaxed about renaming things at the vnode level or above - the v_ bit is redundant.

sys/src/9/ufs/ufs_mountpoint.c
gmacd: Possibly.  I'll add that possibility as a note for later.

sys/src/9/ufs/ufs_mountpoint.c
gmacd: Yes.
Good catch - thanks.

PR #654, opened 2017-09-18
Move code required by libufs into libufs

gmacd:
Needed to move more code into sys/include.  Also split some functions out of ffs_subr into libufs/libufs_subr.c


sys/src/libufs/libufs.json
```diff
@@ -9,6 +9,8 @@
 			"block.c",
 			"cgroup.c",
 			"inode.c",
+			"libufs_subr.c",
+			"ffs_tables.c",
 			"sblock.c",
```
elbing: The standard is that build tool recognize automagicly a file named build.json. I know you saw older format in what we used libfoo.json or cmd.json.

gmacd: What do we call the kernel build.json when we already have build.json?

PR #655, opened 2017-09-19
Fix up some included in libufs

gmacd:

### DIFF
--- a/sys/src/libufs/sblock.c
+++ b/sys/src/libufs/sblock.c
@@ -25,17 +25,16 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
-#include <sys/cdefs.h>
-__FBSDID("$FreeBSD$");
+//#include <sys/cdefs.h>
 
-#include <sys/param.h>
-#include <sys/mount.h>
-#include <sys/disklabel.h>
-#include <sys/stat.h>
+//#include <sys/param.h>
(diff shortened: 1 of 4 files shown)


sys/src/libufs/sblock.c
```diff
@@ -25,25 +25,24 @@
...
-#include <stdlib.h>
-#include <unistd.h>
+//#include <errno.h>
+//#include <stdio.h>
+//#include <string.h>
+//#include <stdlib.h>
+//#include <unistd.h>
 
-#include <libufs.h>
+#include <ufs/libufs.h>
```
rminnich: ufs/libufs.h here

sys/src/libufs/type.c
```diff
@@ -25,28 +25,27 @@
...
+//#include <errno.h>
+//#include <fcntl.h>
+//#include <fstab.h>
+//#include <paths.h>
+//#include <stdio.h>
+//#include <stdlib.h>
+//#include <string.h>
+//#include <unistd.h>
+
+#include <lib/libufs.h>
```
rminnich: lib/libufs.h here

this is a bit confusing. DO they need to be separate (maybe they do)?)

gmacd: Nope, typo.  Was meant to be <ufs/libufs.h>.

PR #664, opened 2017-10-03
newfs - handle all valid args

gmacd:


rminnich:
most of those optimizations probably go all the way back to RP03 days, and they're meaningless now -- the drives do all kinds of things to your IO requests. There's a full RTOS on the drive and it knows way better what to do than you do.

PR #666, opened 2017-10-19
Remove all LITTLE_ENDIAN tests from ufs code

gmacd:
By assuming that we only handle 'new' file systems with a fs_maxsymlinklen > 0, we can avoid all the LITTLE_ENDIAN tests.

### DIFF
--- a/sys/include/ufs/dir.h
+++ b/sys/include/ufs/dir.h
@@ -104,21 +104,12 @@ typedef struct Direct {
  * the directory entry.  This requires the amount of space in struct direct
  * without the d_name field, plus enough space for the name with a terminating
  * null byte (dp->d_namlen+1), rounded up to a 4 byte boundary.
- *
- * 
  */
 #define	DIRECTSIZ(namlen)						\
 	((offsetof(Direct, d_name) +				\
 	  ((namlen)+1)*sizeof(((Direct *)0)->d_name[0]) + 3) & ~3)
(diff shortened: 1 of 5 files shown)


sys/include/ufs/dir.h
```diff
@@ -104,21 +104,12 @@ typedef struct Direct {
...
 #define	DIRECTSIZ(namlen)						\
 	((offsetof(Direct, d_name) +				\
 	  ((namlen)+1)*sizeof(((Direct *)0)->d_name[0]) + 3) & ~3)
-#if (BYTE_ORDER == LITTLE_ENDIAN)
-#define	DIRSIZ(oldfmt, dp) \
-    ((oldfmt) ? DIRECTSIZ((dp)->d_type) : DIRECTSIZ((dp)->d_namlen))
-#else
-#define	DIRSIZ(oldfmt, dp) \
+#define	DIRSIZ(dp) \
     DIRECTSIZ((dp)->d_namlen)
```
rminnich: so will this now break on bigendian machines?

gmacd: No.  All the calls pass oldfmt as 0, or call a macro that returns 0 if the file system is not an old-style filesystem (by checking the mnt_maxsymlinklen).  All our filesystems will be created with that value > 0, since we don't support UFS1.  So if we assume oldfmt is always 0, the big and little endian code is exactly the same, so we don't need the ifdef.

PR #670, opened 2017-10-24
Remove some more UFS1 code

gmacd:
Also remove some duplicate code.  Found while examining the superblock.

### DIFF
--- a/sys/include/ffs/fs.h
+++ b/sys/include/ffs/fs.h
@@ -239,14 +239,14 @@ typedef struct csum {
 	int32_t	cs_nifree;		/* number of free inodes */
 	int32_t	cs_nffree;		/* number of free frags */
 } csum;
-struct csum_total {
+typedef struct csum_total {
 	int64_t	cs_ndir;		/* number of directories */
 	int64_t	cs_nbfree;		/* number of free blocks */
 	int64_t	cs_nifree;		/* number of free inodes */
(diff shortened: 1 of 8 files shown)


sys/include/ffs/fs.h
```diff
@@ -511,9 +511,7 @@ typedef struct Cg {
 #define	cgimin(fs, c)	(cgstart(fs, c) + (fs)->fs_iblkno)	/* inode blk */
 #define	cgsblock(fs, c)	(cgstart(fs, c) + (fs)->fs_sblkno)	/* super blk */
 #define	cgtod(fs, c)	(cgstart(fs, c) + (fs)->fs_cblkno)	/* cg block */
-#define	cgstart(fs, c)							\
-       ((fs)->fs_magic == FS_UFS2_MAGIC ? cgbase(fs, c) :		\
-       (cgbase(fs, c) + (fs)->fs_old_cgoffset * ((c) & ~((fs)->fs_old_cgmask))))
+#define	cgstart(fs, c)	(cgbase(fs, c))
```
elbing: this is not UFS1_MAGIC, it's UFS2

gmacd: For our file system, fs_magic will be FS_UFS2_MAGIC, so we can just use the first expression (cgbase(fs, c))

PR #673, opened 2017-11-04
Move UFS files into a smaller number of directories

gmacd:
A little spring cleaning before doing some more interesting work on vnodes/inodes.

Move headers into a single ufs folder.
Also move ufs .c files into a single ufs folder.


gmacd:
That should be a little tidier now.

PR #675, opened 2017-11-05
Display inode details

gmacd:
Also, I shouldn't have added lrand() to the kerel lib.h.  This function uses a different Lock to that in the kernel (but with same name), and ths caused problems.

### DIFF
--- a/sys/src/9/ufs/ffs_vfsops.c
+++ b/sys/src/9/ufs/ffs_vfsops.c
@@ -1633,7 +1633,7 @@ ffs_vgetf(MountPoint *mp, ino_t ino, int flags, vnode **vpp, int ffs_flags)
 			// lrand() returns int32_t, which isn't ideal but it's
 			// all we have right now
 			// TODO HARVEY We need a u64 random number generator
-			ip->i_gen = (int)-INT32_MIN + rand();
+			ip->i_gen = (int64_t)-INT32_MIN + rand();
 		}
 		if ((vp->mount->mnt_flag & MNT_RDONLY) == 0) {
 			ip->i_flag |= IN_MODIFIED;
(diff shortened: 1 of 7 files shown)


sys/src/9/ufs/ffs_vfsops.c
```diff
@@ -1629,7 +1632,8 @@ ffs_vgetf(MountPoint *mp, ino_t ino, int flags, vnode **vpp, int ffs_flags)
 			// previously the source here), returns uint32_t.
 			// lrand() returns int32_t, which isn't ideal but it's
 			// all we have right now
-			ip->i_gen = lrand();
+			// TODO HARVEY We need a u64 random number generator
+			ip->i_gen = -INT32_MIN + rand();
```
elbing: Talk with Rafa (@fuchicar) for it. I think he included it when made the actual random device removing the old from plan9.

elbing:
Seems old gcc's don't like:
../ufs/ffs_vfsops.c:1636:16: error: integer overflow in expression [-Werror=overflow]

sys/src/9/ufs/ffs_vfsops.c
```diff
@@ -1633,7 +1633,7 @@ ffs_vgetf(MountPoint *mp, ino_t ino, int flags, vnode **vpp, int ffs_flags)
 			// lrand() returns int32_t, which isn't ideal but it's
 			// all we have right now
 			// TODO HARVEY We need a u64 random number generator
-			ip->i_gen = -INT32_MIN + rand();
+			ip->i_gen = (int)-INT32_MIN + rand();
```
elbing: int64_t

gmacd: Yeah, you're right. Also, it shouldn't be 0 at all, and that seems to be the case in my testing, so I need to fix something in newfs.

Issue #605, opened 2017-05-22
Replace TAILQ_HEAD, LIST_HEAD queue and list macros in libufs


gmacd:
From PR #604: 'These are kinda nasty and definitely not plan 9 style. Is there some way to adapt the code to not use these? Or change them to look like plan 9 code?'

Once we have libufs compiling and working, we can address this.

PR #679, opened 2017-11-14
Another UFS tidying up commit

gmacd:
Trying to consolidate harvey-specific code into sensible .c and .h files.
Reading is in progress, this is more prep work.

### DIFF
--- a/sys/include/ufs/inode.h
+++ b/sys/include/ufs/inode.h
@@ -59,7 +59,7 @@ typedef struct ufs2_dinode ufs2_dinode;
  * exclusive.
  */
 struct inode {
-	TAILQ_ENTRY(inode) i_nextsnap; /* snapshot file list. */
+	//TAILQ_ENTRY(inode) i_nextsnap; /* snapshot file list. */
 	struct	vnode  *i_vnode;/* Vnode associated with this inode. */
 	struct 	ufsmount *i_ump;/* Ufsmount point associated with this inode. */
 	struct	 dquot *i_dquot[MAXQUOTAS]; /* Dquot structures. */
--- a/sys/include/ufs/inode.h
+++ b/sys/include/ufs/inode.h
@@ -59,6 +59,7 @@ typedef struct ufs2_dinode ufs2_dinode;
  * exclusive.
  */
 struct inode {
+	// TODO HARVEY Replace when snapshot code re-enabled
 	//TAILQ_ENTRY(inode) i_nextsnap; /* snapshot file list. */
 	struct	vnode  *i_vnode;/* Vnode associated with this inode. */
 	struct 	ufsmount *i_ump;/* Ufsmount point associated with this inode. */
--- a/sys/src/9/ufs/ffs_vfsops.c
+++ b/sys/src/9/ufs/ffs_vfsops.c
@@ -1506,9 +1506,9 @@ ffs_vgetf(MountPoint *mp, ino_t ino, int flags, vnode **vpp, int ffs_flags)
 	vnode *vp;
 	int error;
 
-	error = findexistingvnode(mp, ino, vpp);
-	if (error || *vpp != nil)
-		return (error);
+	*vpp = findvnode(mp, ino);
+	if (*vpp != nil)
+		return 0;
 
 	/*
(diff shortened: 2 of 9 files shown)


sys/src/9/ufs/ffs_vfsops.c
```diff
@@ -1591,7 +1591,7 @@ ffs_vgetf(MountPoint *mp, ino_t ino, int flags, vnode **vpp, int ffs_flags)
 		 * list by vput().
 		 */
 		free(buf);
-		releaseufsvnode(vp);
+		releasevnode(vp);
 		*vpp = nil;
 		return (error);
```
rminnich: why the ()?

sys/src/9/ufs/ufs_harvey_vnode.c
```diff
@@ -117,41 +195,40 @@ getnewvnode(MountPoint *mp, vnode **vpp)
...
+		vn->prev = nil;
 	}
-	return vn;
+
+	qunlock(&mp->vnodes_lock);
 }
 
 int
-lookuppath(MountPoint *mp, char *path, vnode **vn)
+countvnodes(vnode* vn)
```
rminnich: If this is called with something locked you might want to say so. If there is no lock of any kind should we worry?

sys/src/9/ufs/ufsfns.h
```diff
@@ -14,36 +14,30 @@ typedef struct MountPoint MountPoint;
...
-
-void assert_vop_locked(vnode *vp, const char *str);
-void assert_vop_elocked(vnode *vp, const char *str);
-
-int32_t bread(MountPoint *mp, ufs2_daddr_t blockno, size_t size, void **buf);
-
-vnode* findvnode(MountPoint *mp, ino_t ino);
-vnode* getfreevnode(MountPoint *mp);
-void releasevnode(vnode *vn);
+// Vnode
```
rminnich: ufsnfs.h?

special stuff for nfs?

sys/include/ufs/inode.h
```diff
@@ -59,7 +59,7 @@ typedef struct ufs2_dinode ufs2_dinode;
  * exclusive.
  */
 struct inode {
-	TAILQ_ENTRY(inode) i_nextsnap; /* snapshot file list. */
+	//TAILQ_ENTRY(inode) i_nextsnap; /* snapshot file list. */
```
rminnich: will this ever be used again? If it might get turned on again maybe a TODO is a good idea? If never again, remove it?

gmacd: It'll be used.  When we enable the code that uses it, it'll be clear that it's needed.  I can add a todo though, just in case anyone else removes it.

sys/src/9/ufs/ufsfns.h
gmacd: ufsfns - ufs functions.

sys/include/ufs/inode.h
elbing: We have an opened issue for tailq, references could be there and mentioned in a comment.

sys/src/9/ufs/ufs_harvey_vnode.c
gmacd: It assumes the vnodes have been locked.  I should add a check though.

sys/src/9/ufs/ffs_vfsops.c
elbing: Seems it came with original. There was an age in what everybody used parenthesis in return statements.

gmacd: Ah, I was going to come back to this - I didn't follow the question.  If you're referring to the return statement, then Alvaro's right - it's the original code.

sys/include/ufs/inode.h
gmacd: The intention is to replace it with non-macro lists, as in issue #605

gmacd:
Let me know if you're happy with the changes and I'll push.

sys/src/9/ufs/ufsfns.h
rminnich: I blame my glasses.

PR #681, opened 2017-11-20
Adding jpg and page programs

elbing:
We can now read, show and convert some image formats with this programs.
TODO: pdf are not available, it needs ghostscript in a next round.


sevki:
should this be renamed to `cmd/image` instead of `cmd/jpg` since you have `gif` and `png` there as well?

elbing:
Not really since every program has its own flags and behaviour:
http://man.9front.org/1/jpg

sevki:
I don't get it but I'll take your word for it

PR #683, opened 2017-11-25
ufs_readdir + Buf and Uio implementations

gmacd:
Quite a lot here, so apologies, but one thing led to another...

* Implemented ufs_readdir.  This can be tested by mounting a fs, and calling ```cat /dev/ufs/0/inodes/n/data``` where n is the inode of a directory.
* This required a first pass at implementing Buf and Uio, so as to minimise changes needed in UFS.  It's really only taking reading into account.  It'll need another pass (or several) when we get to writing.
* Uio uses the existing plan 9 Block api.  I was wondering whether I could use Queue, but haven't needed to yet.
* Enabled a good chunk of ufs_bmap


rminnich:
The idea of using Queue ought to work, but you should review the akaros changes (bug fixes) to make sure that are not some we need.

PR #685, opened 2017-11-28
A few work on sys/src/cmd

elbing:
Programs added to build:
tbl, html2roff, troff2html, cfs, cifs, ext2srv, paqfs, proof and scat.

Programs moved to APEX due to dependencies (not ported yet, see there PR):
lp daemons and pic.

Programs moved to legacy repo:
postscript, ratfs, webfs and wikifs.

Remaining in src/cmd:
aquarela
dict
fax
grap
scuzz
snap
spell
spin
tapefs
tcs
vac
vnc

Surely, ip/telnet, ipt/telnetd and their front-ends ssh1 (obsolete protocol) and ssh2 will be moved to legacy because Harvey has deprecated (but not removed) the syscalls segattach, segdettach, segfree and the C func sbrk().


sys/src/cmd/cifs/auth.c
```diff
@@ -63,7 +47,7 @@ auth_plain(char *windom, char *keyp, uint8_t *chal, int len)
 		sysfatal("cannot get key - %r");
 
 	ap = emalloc9p(sizeof(Auth));
-	memset(ap, 0, sizeof(ap));
+	memset(&ap->mackey, 0, sizeof(ap->mackey));
```
gmacd: Why did this change?  Compiler warning?

sys/src/cmd/paqfs/paqfs.c
```diff
@@ -261,14 +261,14 @@ main(int argc, char *argv[])
 		break;
 	default:
 		close(pfd[1]);	/* don't deadlock if child fails */
-		if(mnt && mount(pfd[0], -1, mntpoint, MREPL|MCREATE, "") < 0)
+		if(mnt && (mount(pfd[0], -1, mntpoint, MREPL|MCREATE, "", 'M') < 0))
```
gmacd: What's the 'M' for here?

sys/src/cmd/cifs/auth.c
elbing: deanon structs. Compiler tells you what it wants.

sys/src/cmd/paqfs/paqfs.c
elbing: It's not in man, that flag was added by Ron to mount at early days (I think after an agreement with John and Aki, they were the gurus). 
```
commit 72777a1fab7b33c26bb4d9dccff934d4f213303e
Author: Ronald G. Minnich <rminnich@gmail.com>
Date:   Thu Aug 20 09:49:28 2015 -0700

    Expose the 'dc' parameter to user programs by extending the mount call.
    
    The mount call has another parameter, an int, which is the device character
    used to pick a mount device. This works for mount and ramfs,
    but it would be nice if someone else could verify it.
    
    Now, go forth, and write new devmnt devices!
    
    Change-Id: I765dd6830f5a26d8a96d5e568fe61871cf7e402c
```
Usually, #M mnt is the only one used, that's why 99,9% of mount() calls you can see in src/cmd have 'M', but Ron made #N mntn for showing how to make more devmnt devices. The main purpose was making devmnt's for new drivers or fs that couldn't be necessarily 9P or that could have special requirements, like ZFS or ext4 as we thought those days.

PR #687, opened 2017-11-29
Enable reading of files in UFS

gmacd:
cat /dev/ufs/0/inodes/<fileinode>/data

Also handle file offsets better when dumping files or directories.
Quite a few todos here.


gmacd:
Yeah, next up is path lookup :)

PR #688, opened 2017-11-30
Removing ssh/telnet/rlogin services due to old syscall dependency

elbing:


elbing:
There are all telnetd in fact and it's based on segattach which doesn't work and is deprecated. I talked with Ron after trying to making work ssh2 (it worked through rsa key until calling telnetd) and is cheaper to port openssh. Who cares telnet, you have an insecure console at the end of cpurc for that, just telnet.go needs a few fix since we put back cons in kernel.

PR #689, opened 2017-12-01
Adding scuzz, snap and vnc, removing obsolete spelling/dictionary tools and fax (no telco)

elbing:


elbing:
This leaves pending of being ported:
```
aquarela -> CIFS/Samba server
fax -> Fax receiver/sender fs
grap -> pic(1) preprocessor for drawing graphs (another one)
snap -> save/restore static process images fs
spin -> verification tool for models of concurrent systems (historic, maybe interesting)
tapefs -> tape storing system fs
tcs -> character set translator (classic unix program)
vac -> archival copy of trees for venti (harvey has venti)
vnc -> server/viewer remote frame buffer
```

elbing:
ok?

sys/src/cmd/snap/take.c
```diff
@@ -91,7 +91,7 @@ readsection(int32_t pid, char *sec)
 		return nil;
 
 	tot = 0;
-	hdr = (int)((Data*)0)->data;
+	hdr = (int)(intptr_t)((Data*)0)->data;
```
gmacd: So it all looks good, but I don't understand this - is this not dereferencing a null pointer?  It's late and I'm tired, so maybe I've missed something...  (I realise it was the same in the original code)

elbing: no no, you're in right. I hadn't a better idea for this case. It's a case of assignement of smaller integer than de pointer hdr, I think. Let me review it tomorrow in the machine in what I have the branch and I'll paste the right warning/error. Maybe you could have a better solution, I wasn't sure about this and it's ugly. Thanks!

gmacd: You're not making it any worse, so feel free to merge the PR.

rminnich: you might want to use offsetof
#define offsetof(st, m) ((size_t)&(((st *)0)->m))
You're missing an &

rminnich: I think you should fix this before you commit.

elbing: The problem is not with clang, is with gcc:
```
take.c: In function ‘readsection’:
take.c:94:8: error: cast from pointer to integer of different size [-Werror=pointer-to-int-cast]
  hdr = (int)((Data*)0)->data;
```
Not sure why clang eats it.

elbing: Thanks, Ron. It works like a charm:
```
	tot = 0;
	hdr = offsetof(Data, data);
	while((n = read(fd, buf, sizeof buf)) > 0) {
```
libc.h always to the rescue!

sys/src/cmd/vnc/compat.c
```diff
@@ -73,7 +73,11 @@ void
 initcompat(void)
 {
 	rfork(RFREND);
-	privup = privalloc();
+	/* I just don't know how to fix this, because
+	 * in harvey up is Mach *mah()->externup and
+	 * this emulation through the proc seems so weird for me.
+	 */
+	privup = (Proc **)privalloc();
 	kerndate = seconds();
```
elbing: suggestions?

elbing:
Sorry for adding vnc, it was in the same branch. I pointed a doubt in `sys/src/cmd/vnc/compat.c`

Issue #698, opened 2018-01-02
Fix Copyrights


Shamar:
After the recent reintroduction of the missing 9front's copyright statements to the libsec's code at 81d25a4063b3ae76e993fbab3a06661a2ca68e00, it has been jokingly pointed to me that "I'm not worthy", since [none of my 6 months of commits](https://github.com/Harvey-OS/harvey/commits?author=Shamar) have left a sign in Harvey. :-)

After a grep confirming the fact, I've noticed that my port of zlib has been removed and the console has been put back in kernel (removing the user-space filesystem that I wrote in 2015). Also I know that my contributions to the continuous integration and the Coverity integration cannot be evident into the sources.

However looking around I've noticed that at 066eb48ac5025570bf5a63a8ef24da4b35f4cbc2 **my name and copyright has been removed from two new files** that, despite simple in itself, were part of two non trivial contributions at cf801d01f503fc340efd6d0ca97a4921b24b447c and 5616a46061a3a6ea84375e0c34dc8c9dfe9ee87c that moved read(2) and write(2) to user-space while cleaning up the kernel. I realize now that I should have probably amended the copyright of all the files touched by the change, to avoid filling this issue now.

Since I can't find the AUTHORS file referenced by @rminnich in the commit message at 066eb48ac5025570bf5a63a8ef24da4b35f4cbc2 it's clear that, actually, an AUTHORS file cannot replace proper copyright statements in the code (as [wisely predicted by Ron himself in the mailing list](https://groups.google.com/d/msg/harvey/aG4f6ABDJAw/T3AprBwHDQAJ)).

I do not remember if I have contributed other non trivial changes to Harvey (that have not been reverted or removed). For example I do not know if the changes at 401284bef0d6ae6d240e1d27dece8a5165011316 or at 4392ad33cbf6ab719c02c535158b01014d17ec69 were trivial or I had to add my copyright to the files modified. I was pretty naive back then, but I guess that if the change is not trivial, the copyright holds even if not noticed in the license's header, so it would probably be better to make it explicit.

Nevertheless, since my name has gone in libc, **it's possible that it happened to other contributors too** and while I'm not a lawyer, I guess this should be fixed. Unfortunately I have no time to check them all.

As for me I prefer the copyright to be put in the files I contributed to, [as suggested by the Ron back in 2015](https://groups.google.com/d/msg/harvey/aG4f6ABDJAw/T3AprBwHDQAJ).  
[...]

PR #699, opened 2018-01-03
Fixing Copyright in libc write.c and read.c

elbing:


elbing:
Issue #698 .

The agreement since Harvey entered SFC is that no any personal copyright is allowed in files originally under the GPLv2 tree, just in those added by authors and no part of original tree or those imported from other projects already licensed and if their license is compatible with the GPL. That's why we removed all our copyrights, replacing them for Harvey OS which is the name of the project.

Those two files mentioned in issue #698 had already a previous copyright, we restore them as author requested us.

If there's any other file added to harvey that was previously copyrighted and was removed due to the later agreement, please warn us.

All authors are in historic log of the master repo and we appreciate all their efforts and dedication to Harvey.

Thanks

rminnich:
Yeah, my original guidance on this was pre-SFC, it's now considered best to not put more copyright notices in a file. That's what we do on u-root: no names in files.
I regret any confusion I created.

Issue #690, opened 2017-12-02
ratrace suicide if run with no args


gmacd:
#### Expected behavior
Usage info if being run with no args is invalid.  No suicide.

#### Actual behavior
Suicide

#### Steps to reproduce
- Run 'ratrace'
- See it die

#### QEMU output
```
read fault fail, no segment, pid 91 addr 0x0 pc 0x40085b
```

elbing:
#782 by @keedon

PR #717, opened 2018-01-16
Fixed suicide in ratrace when no arguments are passed

keedon:
Signed-off-by: Keith Poole <keef@localhost.localdomain>


elbing:
This is for solving #690 issue, but needs to be done against newer master. When you can :-).

elbing:
will be a new pr.

PR #718, opened 2018-01-17
Importing read and write from APEX

elbing:
Bringing up wirte and read from ap/lib/_write.c and ap/lib/_read.c
  in APEX.

### DIFF
--- a/amd64/include/u.h
+++ b/amd64/include/u.h
@@ -15,6 +15,7 @@ typedef	signed short	int16_t;
 typedef unsigned int	uint32_t;
 typedef unsigned int	uint;
 typedef signed int	int32_t;
+typedef long		ssize_t;
 typedef	unsigned long long uint64_t;
 typedef	long long	int64_t;
 typedef uint64_t uintptr;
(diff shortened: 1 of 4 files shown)


amd64/include/u.h
```diff
@@ -15,6 +15,7 @@ typedef	signed short	int16_t;
 typedef unsigned int	uint32_t;
 typedef unsigned int	uint;
 typedef signed int	int32_t;
+typedef long		ssize_t;
```
gmacd: Should ssize_t be long by default?  I understand it should be in for this import, but in the general case?

gmacd: Seems to be int64 in Freebsd.

elbing: Gcc manages long as int64 in amd64 arch. I dealed with it in aex many times. See size_t for unsigned long. Ron could explain it better. 32 bit integer is int and int32_t. We changed long for int32 because most of the types in plan9 were designed for 32 bit arch and a long is an int32_t in i386 arch. In harvey we don't use size and ssize natively. This is a special case since those sources are imported from ansi world.

gmacd: Ok, thanks.  If we're bringing the functions out of apex, it would be nice if the signature types matched pread/pwrite.  Anyway, I'll approve and try it out later.

PR #743, opened 2018-01-24
Fixing new imports from APEX

elbing:


sys/src/libc/klibc.json
```diff
@@ -56,6 +57,7 @@
 			"9sys/waitpid.c",
 			"9sys/werrstr.c",
 			"9sys/writev.c",
+			"9sys/writev.c",
```
gmacd: Is this not the same thing included twice?

elbing:
Changed, thanks!

PR #753, opened 2018-01-25
Fix more cmd errors - exportfs, ftpd

gmacd:

### DIFF
--- a/sys/src/cmd/exportfs/exportfs.c
+++ b/sys/src/cmd/exportfs/exportfs.c
@@ -184,6 +184,7 @@ main(int argc, char **argv)
 	default:
 		usage();
 	}ARGEND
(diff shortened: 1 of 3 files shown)


sys/src/cmd/exportfs/exportfs.c
```diff
@@ -184,7 +184,6 @@ main(int argc, char **argv)
 	default:
 		usage();
 	}ARGEND
-	USED(argc, argv);
```
elbing: USED(argc); USED(argv);

gmacd: Seemed good enough for clang - I guess gcc has to be fussy :)

gmacd:
Ok, made the fix.

PR #757, opened 2018-01-28
Libmemdraw

elbing:


gmacd:
I think only the last of these commits needs to be applied - weren't the first two done already?  New commit looks good.

elbing:
Fixed, thanks.

PR #758, opened 2018-01-28
Fixed silly bug in devramfs - added test (/usr/harvey/bin/rc/ramdisk)

keedon:
Just leaving this here until we can have pull requests again.

Signed-off-by: keith.poole <keith.poole@gmail.com>

### DIFF
--- a/sys/src/9/port/devramfs.c
+++ b/sys/src/9/port/devramfs.c
@@ -8,13 +8,13 @@
  */
 
 #include	"u.h"
-#include       "../port/lib.h"
-#include       "mem.h"
-#include       "dat.h"
-#include       "fns.h"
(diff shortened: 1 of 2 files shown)


gmacd:
A couple of random things:
 - What was the bug?  There's a lot of changes here, so hard to pinpoint the bug fix.
 - Maybe we should have a folder for test scripts rather than having them in a common folder (I need the same for UFS test scripts at some point)

keedon:
the actual bug fix is line 140 "if(current->perm & QTDIR){" is now "if(current->perm & DMDIR){"

elbing:
This can be gone isn't it?

elbing:
done.

PR #759, opened 2018-01-30
Various fixes:

elbing:
- assert doesn't need to be restrict to const.
- builtin_trap is an alternative for abort under certain circumstances,
  our need is the original loop.
- aesthetic fix in mksys.go for boot methods


elbing:
This fixes assert issue, but it doesn't boot due to exec() issue.

PR #763, opened 2018-02-02
Back to 2015 in sysfile.c for pread and pwrite

elbing:

### DIFF
--- a/sys/src/9/port/syscallfmt.c
+++ b/sys/src/9/port/syscallfmt.c
@@ -264,20 +264,25 @@ syscallfmt(uint8_t what, int syscallno, Ar0 *ar0, uint64_t start, uint64_t stop,
 		fmtprint(&fmt, "%d ", i[0]);
 		v = va_arg(list, void*);
 		l = va_arg(list, int32_t);
-		vl = va_arg(list, int64_t);
-		if (what == 'E') {
-			fmtprint(&fmt, "%#P %ld 0x%llx", v, l, vl);
+		fmtprint(&fmt, "%d %#p %ld", i[0], v, l);
+		if(syscallno == PREAD){
+			vl = va_arg(list, int64_t);
+			if (what == 'E') {
+				fmtprint(&fmt, "%#P %ld 0x%llx", v, l, vl);
+			}
 		}
 		break;
 	case PWRITE:
 		i[0] = va_arg(list, int);
 		v = va_arg(list, void*);
 		l = va_arg(list, int32_t);
-		vl = va_arg(list, int64_t);
 		fmtprint(&fmt, "%d ", i[0]);
 		len = MIN(l, 64);
-		fmtrwdata(&fmt, v, len, " ");
-		fmtprint(&fmt, "%ld 0x%llx", l, vl);
+		fmtprint(&fmt, "%ld", l);
+		if(syscallno == PWRITE){
+			vl = va_arg(list, int64_t);
+			fmtprint(&fmt, " 0x%llx", vl);
+		}
 		break;
 	}
 	if (what == 'E') {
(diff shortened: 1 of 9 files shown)


sys/src/9/port/syscallfmt.c
```diff
@@ -264,20 +264,25 @@ syscallfmt(uint8_t what, int syscallno, Ar0 *ar0, uint64_t start, uint64_t stop,
 		fmtprint(&fmt, "%d ", i[0]);
 		v = va_arg(list, void*);
 		l = va_arg(list, int32_t);
-		vl = va_arg(list, int64_t);
-		if (what == 'E') {
-			fmtprint(&fmt, "%#P %ld 0x%llx", v, l, vl);
+		fmtprint(&fmt, "%d %#p %ld", i[0], v, l);
+		if(syscallno == PREAD){
```
floren: This is already within a switch on syscallno, in the 'case PREAD' section

sys/src/9/port/syscallfmt.c
```diff
@@ -264,20 +264,25 @@ syscallfmt(uint8_t what, int syscallno, Ar0 *ar0, uint64_t start, uint64_t stop,
...
 		i[0] = va_arg(list, int);
 		v = va_arg(list, void*);
 		l = va_arg(list, int32_t);
-		vl = va_arg(list, int64_t);
 		fmtprint(&fmt, "%d ", i[0]);
 		len = MIN(l, 64);
-		fmtrwdata(&fmt, v, len, " ");
-		fmtprint(&fmt, "%ld 0x%llx", l, vl);
+		fmtprint(&fmt, "%ld", l);
+		if(syscallno == PWRITE){
```
floren: same here

sys/src/9/port/sysfile.c
```diff
@@ -868,6 +882,22 @@ write(int fd, void *p, int32_t n, int64_t off)
...
 
+void
+syswrite(Ar0* ar0, ...)
+{
+	va_list list;
+	va_start(list, ar0);
+	int fd = va_arg(list, int);
+	void *buf = va_arg(list, void *);
+	long nbytes = va_arg(list, long);
+	int64_t offset = -1ULL;
```
rminnich: note that this is right, and the above is wrong, and you don't want to be inconsistent.

sys/src/9/port/sysfile.c
```diff
@@ -793,6 +794,25 @@ read(int fd, void *p, int32_t n, int64_t off)
 	return nnn;
 }
 
+void
+sysread(Ar0* ar0, ...)
+{
+	int fd;
+	void *p;
+	int32_t n;
+	int64_t off = ~0ULL;
```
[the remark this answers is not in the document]
elbing: it can't be -1LL unless the function gets offset arg as a pointer and not as an integer.

sys/src/9/port/sysfile.c
elbing: because in original from NIX and NxM offset is not offp from func arg. I'm not sure why in harvey read and write were different in the sysfile.c imported in 2015.

rminnich: I don't understand

sys/src/9/port/sysfile.c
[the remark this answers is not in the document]
rminnich: what?

sys/src/9/port/syscallfmt.c
rminnich: yeah. The IF was only here back when the case was
case READ:
case PREAD:

sys/src/9/port/syscallfmt.c
rminnich: yeah, this is wrong.

sys/src/9/port/syscallfmt.c
```diff
@@ -353,7 +358,12 @@ syscallfmt(uint8_t what, int syscallno, Ar0 *ar0, uint64_t start, uint64_t stop,
 			fmtprint(&fmt, " %#p/\"\"", v);
 			errstr = up->syserrstr;
 		}
-		fmtprint(&fmt, " %ld 0x%llx = %d", l, vl, ar0->i);
+		fmtprint(&fmt, " %ld", l);
+		if(syscallno == PREAD){
```
rminnich: wrong here too

sys/src/9/port/sysfile.c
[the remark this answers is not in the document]
rminnich: that makes no sense at all.

PR #764, opened 2018-02-02
Re-add missing comparison from last last commit.

gmacd:
Fixes boot issues with cpu.


sys/src/9/port/devmnt.c
```diff
@@ -730,7 +730,7 @@ mntrdwr(int type, Chan *c, void *buf, int32_t n, int64_t off)
 		r->request.offset = off;
 		r->request.data = uba;
 		nr = n;
-		if(mnt->msize-IOHDRSZ)
+		if(nr > mnt->msize-IOHDRSZ)
```
rminnich: good catch.

PR #770, opened 2018-02-04
Delete the remaining unneeded 386 files

gmacd:


gmacd:
I didn't realise there was anything left to delete.

So (in addition to the 386/* files) for upas we have:
./sys/man/4/upasfs
./cron/upas
./rc/bin/startupasfs

I don't see any other bitsy files (at least not anything named bitsy)

Lots of mkfiles I can delete.

The only psh files I see remaining are in apex.

A search for troff gives me:
./amd64/bin/troff2html
./sys/lib/postscript/troff
./sys/lib/troff
./sys/doc/troff.ms
./sys/doc/troff.ps
./sys/doc/troff.pdf
./sys/src/cmd/pic/pltroff.c
./sys/src/cmd/troff2html
./sys/src/cmd/troff2html/troff2html.c
./sys/src/cmd/proof/htroff.c
./sys/man/1/troff2html
./sys/man/1/troff
./rc/bin/troff2png
./rc/bin/troff2gif

gmacd:
If the above files are all deletable let me know and I'll tidy up the commits.

BTW, I don't quite follow 'but ONLY if this is not in /rc/bin.'  You want to leave the rc/bin files in the above lists?

elbing:
All old APE stuff have to be in APEX repo, even if it's an rc script, APEX has its own rc dir binded properly in user profile in case the user has APEX available. Harvey only should have native code. I hope this could solve part of the debate. Also the man pages of APEX programs can be moved to APEX man dir, for now inexistent, but suitable to be binded in the same way in case we need it.
About the order of removing process for deprecated software... for me doesn't matter, it will be gone anyway.

gmacd:
So @rminnich - do you want me to leave the following files in place?
./rc/bin/startupasfs
./rc/bin/troff2png
./rc/bin/troff2gif

gmacd:
Ok, there's more deleted files.  There's a PR to move the troff files into apex.

PR #771, opened 2018-02-05
Fixing a few things for read/write syscalls and an uneeded abort decl…

elbing:
…aration

### DIFF
--- a/sys/src/9/port/sysfile.c
+++ b/sys/src/9/port/sysfile.c
@@ -717,11 +717,10 @@ mountfix(Chan *c, uint8_t *op, int32_t n, int32_t maxn)
 }
 
 static int32_t
-read(int fd, void *p, int32_t n, int64_t off)
+read(int ispread, int fd, void *p, int32_t n, int64_t off)
 {
 	Proc *up = externup();
 	int32_t nn, nnn;
-	int sequential;
 	Chan *c;
 
 	p = validaddr(p, n, 1);
@@ -742,21 +741,24 @@ read(int fd, void *p, int32_t n, int64_t off)
 	 * The number of bytes read on this fd (c->offset) may be different
 	 * due to rewritings in mountfix.
 	 */
+	if(ispread){	/* use and maintain channel's offset */
 		if(off == -1LL){	/* use and maintain channel's offset */
 			off = c->offset;
-		sequential = 1;
-	} else {
-		sequential = 0;
+			ispread = 0;
+		}
 	}
+	else
+			off = c->offset;
+
 	if(c->qid.type & QTDIR){
 		/*
 		 * Directory read:
 		 * rewind to the beginning of the file if necessary;
 		 * try to fill the buffer via mountrockread;
-		 * set sequential to always maintain the Chan offset.
+		 * clear ispread to always maintain the Chan offset.
 		 */
-		if(off == 0LL){
-			if(sequential){
+		if(off == -1LL){
+			if(!ispread){
 				c->offset = 0;
 				c->devoffset = 0;
 			}
@@ -775,12 +777,12 @@ read(int fd, void *p, int32_t n, int64_t off)
 		}
 		nnn = mountfix(c, p, nn, n);
 
-		sequential = 1;
+		ispread = 0;
 	}
 	else
 		nnn = nn = c->dev->read(c, p, n, off);
 
-	if(sequential){
+	if(!ispread){
 		lock(&c->r.l);
 		c->devoffset += nn;
 		c->offset += nnn;
@@ -810,15 +812,14 @@ syspread(Ar0* ar0, ...)
 	/*
 	 * long pread(int fd, void* buf, long nbytes, int64_t offset);
 	 */
-	ar0->l = read(fd, p, n, off);
+	ar0->l = read(1, fd, p, n, off);
 }
 
 static int32_t
-write(int fd, void *p, int32_t n, int64_t off)
+write(int fd, void *p, int32_t n, int64_t off, int ispwrite)
 {
 	Proc *up = externup();
 	int32_t r;
-	int sequential;
 	Chan *c;
 
 	r = n;
@@ -827,13 +828,8 @@ write(int fd, void *p, int32_t n, int64_t off)
 	n = 0;
 	c = fdtochan(fd, OWRITE, 1, 1);
 
-	if(off == -1LL)
-		sequential = 1;
-	else
-		sequential = 0;
-
 	if(waserror()) {
-		if(sequential){
+		if(!ispwrite){
 			lock(&c->r.l);
 			c->offset -= n;
 			unlock(&c->r.l);
@@ -847,7 +843,7 @@ write(int fd, void *p, int32_t n, int64_t off)
 
 	n = r;
 
-	if(sequential){	/* use and maintain channel's offset */
+	if(off == -1LL){	/* use and maintain channel's offset */
 		lock(&c->r.l);
 		off = c->offset;
 		c->offset += n;
@@ -856,7 +852,7 @@ write(int fd, void *p, int32_t n, int64_t off)
 
 	r = c->dev->write(c, p, n, off);
 
-	if(sequential && r < n){
+	if(!ispwrite && r < n){
 		lock(&c->r.l);
 		c->offset -= n - r;
 		unlock(&c->r.l);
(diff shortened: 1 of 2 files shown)


sys/src/9/port/sysfile.c
```diff
@@ -717,11 +717,10 @@ mountfix(Chan *c, uint8_t *op, int32_t n, int32_t maxn)
 }
 
 static int32_t
-read(int fd, void *p, int32_t n, int64_t off)
+read(int ispread, int fd, void *p, int32_t n, int64_t off)
```
dancrossnyc: Why is this first for read, but last for write?

dancrossnyc: Also, I'd keep the 'sequential' name: it implies `ispread` would be true.

elbing: I just follow the original definition.

sys/src/9/port/sysfile.c
```diff
@@ -742,21 +741,24 @@ read(int fd, void *p, int32_t n, int64_t off)
 	 * The number of bytes read on this fd (c->offset) may be different
 	 * due to rewritings in mountfix.
 	 */
+	if(ispread){	/* use and maintain channel's offset */
```
dancrossnyc: I think you can simplify this to,

    if(sequential || off == -1LL) {
        off = c->offset;
        sequential = 1;
    }

sys/src/9/port/sysfile.c
```diff
@@ -742,21 +741,21 @@ read(int fd, void *p, int32_t n, int64_t off)
...
-		if(off == -1LL){	/* use and maintain channel's offset */
-			off = c->offset;
-		sequential = 1;
-	} else {
-		sequential = 0;
+	if(off == -1LL){	/* use and maintain channel's offset */
+		off = c->offset;
 	}
+	else
+		dopread = 0;
```
dancrossnyc: Isn't this reversed? If off == -1LL, then it's NOT pread (we're relying on the offset tracked in the Chan). If it's pread, then we use the offset passed as an argument.

sys/src/9/port/sysfile.c
```diff
@@ -827,13 +825,11 @@ write(int fd, void *p, int32_t n, int64_t off)
 	n = 0;
 	c = fdtochan(fd, OWRITE, 1, 1);
 
-	if(off == -1LL)
-		sequential = 1;
-	else
-		sequential = 0;
+	if(off != -1LL)
+		dopwrite = 0;
```
dancrossnyc: Same here.

sys/src/9/port/sysfile.c
```diff
@@ -847,7 +843,7 @@ write(int fd, void *p, int32_t n, int64_t off)
 
 	n = r;
 
-	if(sequential){	/* use and maintain channel's offset */
+	if(dopwrite){	/* use and maintain channel's offset */
```
dancrossnyc: Why not avoid advancing c->offset until after r->dev->write returns?

dancrossnyc: Answering my own question...If I had to guess:

The authors increment offset so that multiple system calls writing to the chan at the same time don't overlap. However, there's a time-of-check, time-of-use bug here (I think) with setting `off` when off==-1LL.

Probably the best thing is to add, `int pread = (off != -1LL);` at the beginning and test on that to set off from c->offset and increment c->offset before c->dev->write.

sys/src/9/port/sysfile.c
```diff
@@ -856,7 +852,7 @@ write(int fd, void *p, int32_t n, int64_t off)
 
 	r = c->dev->write(c, p, n, off);
```
dancrossnyc: Hmm. `r` is signed. What if it's -1 here?

util/src/harvey/cmd/elf2c/elf2c.go
```diff
@@ -139,7 +139,7 @@ func main() {
 		gencode(w, file, "code", mem, codestart, codeend)
 		gencode(w, file, "data", mem, datastart, dataend)
 	}
-	if err := ioutil.WriteFile(a[0], w.Bytes(), 0444); err != nil {
+	if err := ioutil.WriteFile(a[0], w.Bytes(), 0666); err != nil {
```
dancrossnyc: Is this related?

elbing: aaaah no, it came from the annoying rm init.h I have to make every time.

sys/src/9/port/sysfile.c
elbing: suicide?

sys/src/9/port/sysfile.c
```diff
@@ -742,21 +741,21 @@ read(int fd, void *p, int32_t n, int64_t off)
...
-		if(off == -1LL){	/* use and maintain channel's offset */
-			off = c->offset;
-		sequential = 1;
-	} else {
-		sequential = 0;
+	if(off == -1LL){	/* use and maintain channel's offset */
+		off = c->offset;
 	}
+	else
+		dopread = 1;
```
dancrossnyc: dopread is already 1?

sys/src/9/port/sysfile.c
```diff
@@ -720,8 +720,7 @@ static int32_t
 read(int fd, void *p, int32_t n, int64_t off)
 {
 	Proc *up = externup();
-	int32_t nn, nnn;
-	int sequential;
+	int32_t nn, nnn, dopread = 1;
```
dancrossnyc: Add a line:

    int dopread = (off != -1LL);

sys/src/9/port/sysfile.c
```diff
@@ -742,21 +741,21 @@ read(int fd, void *p, int32_t n, int64_t off)
 	 * The number of bytes read on this fd (c->offset) may be different
 	 * due to rewritings in mountfix.
 	 */
-		if(off == -1LL){	/* use and maintain channel's offset */
-			off = c->offset;
-		sequential = 1;
-	} else {
-		sequential = 0;
+	if(off == -1LL){	/* use and maintain channel's offset */
```
dancrossnyc: if (!dopread) off = c->offset;

sys/src/9/port/sysfile.c
```diff
@@ -742,21 +741,21 @@ read(int fd, void *p, int32_t n, int64_t off)
...
 		/*
 		 * Directory read:
 		 * rewind to the beginning of the file if necessary;
 		 * try to fill the buffer via mountrockread;
-		 * set sequential to always maintain the Chan offset.
+		 * clear dopread to always maintain the Chan offset.
 		 */
-		if(off == 0LL){
-			if(sequential){
+		if(off == -1LL){
```
dancrossnyc: if(off == 0LL){
        c->offset = 0;
        c->devoffset = 0;
    }
    nnn = mountfix(c, p, nn, n);
    dopread = 0;

sys/src/9/port/sysfile.c
```diff
@@ -775,12 +774,12 @@ read(int fd, void *p, int32_t n, int64_t off)
...
 		nnn = mountfix(c, p, nn, n);
 
-		sequential = 1;
+		dopread = 1;
 	}
 	else
 		nnn = nn = c->dev->read(c, p, n, off);
 
-	if(sequential){
+	if(dopread){
```
dancrossnyc: This is reversed; but I think you can just always update the offsets in every case.

    lock(&c->r.l);
    if(nn > 0)
        c->devoffset += nn;
    if (nnn > 0)
        c->offset += nnn;
    unlock(&c->r.l);

sys/src/9/port/sysfile.c
```diff
@@ -827,13 +825,11 @@ write(int fd, void *p, int32_t n, int64_t off)
...
-	if(off == -1LL)
-		sequential = 1;
-	else
-		sequential = 0;
+	if(off != -1LL)
+		dopwrite = 1;
 
 	if(waserror()) {
-		if(sequential){
+		if(dopwrite){
```
dancrossnyc: if (!doprwrite)

`doprwrite==1` means that you are using `pwrite` semantics. The old `sequential` meant that you were doing `write` semantics.

sys/src/9/port/sysfile.c
```diff
@@ -742,27 +742,24 @@ read(int fd, void *p, int32_t n, int64_t off)
...
 		 */
 		if(off == 0LL){
-			if(sequential){
-				c->offset = 0;
-				c->devoffset = 0;
-			}
-			mountrewind(c);
-			unionrewind(c);
-		}
+ 			if(notpread){
```
dancrossnyc: Looks like some random leading whitespace crept into this change?

elbing: Cleaned!

PR #773, opened 2018-02-08
Fix for libthread - fixes vga/realemu interaction

gmacd:
Undo some changes from ac5494c22127f601346841986a1f80d18a542129
Looks like that commit was meant to fix some compiler errors.
We go back to the original code and fix the errors in a more direct way.


gmacd:
As I typed the commit, I had a similar feeling, but at this stage I'm happy it progresses things.  If we can fix an underlying problem later, then even better.

PR #774, opened 2018-02-12
New README.md

elbing:


gmacd:
Does Travis still work? (Seeing the reference in the new wiki)

elbing:
We'll change wiki when we have a new solution.

PR #781, opened 2018-02-15
Cleaning some messages from boot

elbing:
Those messages now under DBG are useful for implementing
  ACPI or debugging and error. In Qemu there's only a line,
  In real hardware around 2 hundred.
  Same for mmu and vgavesa, not hundreds, but see useless messages
  if it's not debugging time.

### DIFF
--- a/sys/src/9/amd64/mmu.c
+++ b/sys/src/9/amd64/mmu.c
@@ -646,9 +646,9 @@ vmap(uintptr_t pa, usize size)
 	uintptr_t va;
 	usize o, sz;
 
-	DBG("vmap(%#p, %lu) pc=%#p\n", pa, size, getcallerpc());
+	if(DBGFLG) print("vmap(%#p, %lu) pc=%#p\n", pa, size, getcallerpc());
 
-	if(machp()->machno != 0)
+	if(machp()->machno != 0 && DBGFLG)
 		print("vmap: machp()->machno != 0\n");
 
 	/*
@@ -700,7 +700,7 @@ vunmap(void* v, usize size)
 	DBG("vunmap(%#p, %lu)\n", v, size);
 
 	if(machp()->machno != 0)
-		print("vmap: machp()->machno != 0\n");
+		DBG("vmap: machp()->machno != 0\n");
 
 	/*
 	 * See the comments above in vmap.
--- a/sys/src/libacpi/harvey.c
+++ b/sys/src/libacpi/harvey.c
@@ -9,6 +9,9 @@
 
 #include <acpi.h>
 
+#undef DBG
+#define DBG print
+
 #define MiB (1<<20)
 
 BOOLEAN                        AcpiGbl_DebugTimeout = FALSE;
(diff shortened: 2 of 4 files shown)


elbing:
Rebased.

sys/src/9/amd64/mmu.c
```diff
@@ -700,7 +700,7 @@ vunmap(void* v, usize size)
 	DBG("vunmap(%#p, %lu)\n", v, size);
 
 	if(machp()->machno != 0)
-		print("vmap: machp()->machno != 0\n");
+		DBG("vmap: machp()->machno != 0\n");
```
gmacd: Why does this go from print -> DBG, when the change above goes from DBG -> print?  Should they be consistent?

sys/src/9/amd64/mmu.c
```diff
@@ -646,9 +646,9 @@ vmap(uintptr_t pa, usize size)
 	uintptr_t va;
 	usize o, sz;
 
-	DBG("vmap(%#p, %lu) pc=%#p\n", pa, size, getcallerpc());
+	if(DBGFLG) print("vmap(%#p, %lu) pc=%#p\n", pa, size, getcallerpc());
```
gmacd: Why is this changed from DBG?  Isn't it doing the same thing?

sys/src/libacpi/harvey.c
```diff
@@ -113,7 +116,7 @@ ins(uint16_t addr)
 	uint64_t off = addr;
 	uint16_t w;
 	if (pread(acpiio, &w, 2, off) < 2)
-		print("ins(0x%x): %r\n", addr);
+		if(debug) DBG("ins(0x%x): %r\n", addr);
```
gmacd: Might be better to wrap if(debug) in the DBG macro.

sys/src/9/amd64/mmu.c
elbing: Both above have to be in the same way, yes. This is a consecuence of try and error. DBG doesn't avoid the print if I define it as print. DBFLG is used in mmu and if present it doesn't print. There were a DBG for the first vmap doing nothing because DBG wasn't defined. When I defined DBG print, a thousand of lines rised during boot process. That's why I left it as it was and disabled machno!=0 with the flag. I tried to be less intrusive as possible. I think the right way is using if DBFLG as in the rest of file. Without macro.

sys/src/libacpi/harvey.c
elbing: Ron left int debug at the begining for flag use. For gdb or compiling time. I tried to be less intrusive too here. Acpi needs a general clean and a homogeneus way for switching flags for debugging. No more macros. That file is full of "if(debug)".

sys/src/9/amd64/mmu.c
elbing: I, I wrote from my phone. Yes I used DBG in parallel with previous DBG for vunmap

sys/src/9/amd64/mmu.c
elbing: I was printing with DBG, it's defined as a print so it wasn't really being reserved for debugging. That's why I chose a different way.

sys/src/libacpi/harvey.c
elbing: We should have to redo this part with a label or so.

PR #786, opened 2018-02-24
Add regression test for FPU state saving

gmacd:
Child proc sets some xmm registers and goes into a tight loop.  Parent proc parses the fpregs file and checks that the xmm registers have the correct values.

Passes with master, fails if I stop the FPU state from being saved.


bldybuild:
Commit f28bbcc380d4fe5f69011311428b95bef6541dd6 is not signed-off.

Please fix these issues:

>	1. Graham MacDonald has not signed-off f28bbcc.

PR #793, opened 2018-03-18
First attempt at simpler, eager FPU state saving

gmacd:
- Eager saving of state
 - No x87 exception handler
 - No restriction on FPU use in note handlers
 - A few regression tests for the FPU

Missing
 - More tests
 - More modern features (e.g. XSAVE)
 - Probably also some fundamental things missed, but hopefully not too many


elbing:
Mac?

gmacd:
It's there if anyone wants to give it a try 😀

PR #804, opened 2018-04-13
Memdraw fix

elbing:
- Removing unused memsetb and using memset for 8 bit.
- Added a forced little endian conversion by default
  in memfillcolor() (from Plan9 port)


elbing:
This solves the effect in native (didn't happen in drawterm) of minimizing a window and rising it again through winwatch which was corrupting the memory.

PR #826, opened 2018-06-14
Add atomic store and load. Seems we need it

rminnich:


gmacd:
You should also remove the riscv implementations.

PR #828, opened 2018-06-15
WIP: syslinux: add syslinux script.

rminnich:
Very simple first pass, lots to do yet.

### DIFF
--- a/util/syslinux.go
+++ b/util/syslinux.go
@@ -1,0 +1,116 @@
+package main
+
+import (
+	"bytes"
+	"io"
+	"io/ioutil"
+	"log"
+	"os"
+	"os/exec"
+	"path/filepath"
+)
+
+var (
+	setupcmds = []struct {
+		*exec.Cmd
+		in io.Reader
+	}{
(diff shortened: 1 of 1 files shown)


gmacd:
I understand it's WIP, but here are a couple of things I noticed:
 - you need to run it in `util`, otherwise it can't copy the relative files
 - detach() doesn't get called on log.Fatal, so the disk doesn't get unmounted and nbd doesn't get disconnected if there's been an error
 - Consider outputting the command you're running - some errors are a bit vague so it's hard to tell what failed.

gmacd:
I've taken another look, some comments:
- you need to run it in util, otherwise it can't copy the relative files
- detach() doesn't get called on log.Fatal, so the disk doesn't get unmounted and nbd doesn't get disconnected if there's been an error.  This leaves it in an inconsistent state, which you'd see if you came across the above problem
- If you run the script as su, then the harvey.qcow2 file is owned by root, and can't be run unless qemu is also run by root
- The util/GO9PIMG script expects the harvey.qcow2 file to be in $HARVEY, but this script creates it in util, so one of these things should change.

gmacd:
If you can rebase and commit, I can approve and commit my changes after.

PR #829, opened 2018-06-21
Avoid unnecessary retries in sdiahci

gmacd:
We get the retry blk message in iario because ci is cleared between the time it's set and we test it before sleeping.  We don't clear ci manually, but it can be cleared if the ST flag on p->cmd is cleared (and probably for other reasons too).  We clear the st flag at various points throughout the sdiahci code

This fix is a little hacky, but ensures we always sleep at least once, which would always happen in the usual case anyway.

This change also includes a script to build a fossil file system and do a large dircp.  This currently functions as a test, but will become a script to build the plan 9 file system for an image.

### DIFF
--- a/sys/src/9/amd64/sdiahci.c
+++ b/sys/src/9/amd64/sdiahci.c
@@ -202,6 +202,7 @@ struct Ctlr {
 struct Asleep {
 	Aport	*p;
 	int	i;
+	int	slept;
 };
 
 extern SDifc sdiahciifc;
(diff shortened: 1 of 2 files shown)


rminnich:
really nice work on figuring this out.

elbing:
Excellent, Graham.

gmacd:
Thanks. Tbh, I don't think it was serious, but it's good to have worked it out (with help).

PR #842, opened 2018-08-28
Keeping all the tools together

elbing:


gmacd:
Maybe we should delete the SYSLINUX file altogether now - I don't think it's necessary now the script works.

PR #865, opened 2018-10-08
Ensure init.h regenerated every time

gmacd:
(it's write only)

### DIFF
--- a/sys/src/9/amd64/inith.json
+++ b/sys/src/9/amd64/inith.json
@@ -24,6 +24,9 @@
 			"-z",
 			"max-page-size=0x200000"
 		],
+		"Pre": [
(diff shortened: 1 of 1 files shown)


elbing:
Graham, it's not remove due to permissions (you can see it at build time). Check permissions in build.go. We talked about this at any time.

gmacd:
Yeah, I can see.  I assumed it was read-only because it was a generated file (so to avoid people editing it).  Why not remove it as part of build?  I think it was before my time?

gmacd:
To be clear, the way it currently is means that it'll only get regenerated if you do a clean.  So any modifications to init don't get picked up.  So either way I think it should be changed.

PR #866, opened 2018-10-09
Rearrange jsons to cleanly separate kernel and user space configs

gmacd:
Main reason is to allow us to build user space without -mno-implicit-float in clang


gmacd:
This broke the fpu regression tests, so needs a bit more work...

PR #870, opened 2018-10-18
Separate kernel and user build files - user code can now use sse with clang

gmacd:
Separate kernel and user build files to allow us to use different compiler options for kernel and user code.  This means we only use clang's -mno-implicit-float for kernel, fixing floats in varargs for user code.  This makes the vseprint test pass.

A few other things:

- Make clang align stack to 16 bytes for user code (fix GP exception in code that uses some SSE instructions that expect 16 byte alignment)
- Change root build order so less important things that aren't dependencies get built first
- Remove -mno-red-zone flag for user code
- Delete some unused files


gmacd:
Running tests/runall passes

PR #872, opened 2018-10-18
TETS!  Add qemu dependency to travis

gmacd:
Let's see if we can run tests in travis...
Trying to get dependencies first...


elbing:
`The command "sudo apt-get install -y --force-yes $CC rc golang qemu" failed and exited with 100 during `
Use apt, there's no need now for old front ends. Also seems that it has some troubles with libc6 and binutils version, we'll see...

gmacd:
How would you change it? I'm not that familiar with apt. Also, which frontends to drop?

elbing:
Closed due to travis migration (can't be any active repo in legacy platform). We can reopen it later.

PR #873, opened 2018-10-18
Adding an rc front script for shutdown

elbing:

### DIFF
--- a/rc/bin/shutdown
+++ b/rc/bin/shutdown
@@ -1,0 +1,17 @@
+#!/bin/rc
+
+# This script will implement all needed for proper shutdown of the machine
+
+flagfmt='A,C,F,H,L,R,U,a,c,d,f,i,h,l,p,r,s,t,u,1'
+args='[file ...]'
+
+if(! ifs=() eval `{aux/getflags $*}){
+        aux/usage
+        exit usage
+}
+
+if(~ $flagr 1) echo Try "reboot" && exit
+
+echo Shutting down the system...
+sleep 3
+acpi/shutdown $* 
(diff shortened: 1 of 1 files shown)


elbing:
Fshalt doesn't work for now. We need an stable fs distro for that. It hasn't sense in actual crosscompiling linux dependent situation. Also, it shouldn't call fshalt, it should replace it syncing the fs, stopping main processes and then calling acpi shutdown. We'll see.

elbing:
The flags catch all the -h -r now etc... That any unix user could use, avoiding silly issues. Also, they can be used in future for specific shutdown circumstances.

gmacd:
Why not just list the flags we actually use?

elbing:
Ok, close it. I have to do 4K regression :-D. If any user asks for shuttingdown, answer him yourself.

gmacd:
I haven't been able to look at this deeply. Is there any reason to specify flags we don't use here?

gmacd:
Sorry, I was out having a drink. Very tardy! I think for these things at should only declare what we use otherwise leave the flags out. Leave future flags for future alvaro and future graham to deal with😁

gmacd:
I think we should add this command, and add any syncing later.

I don't think the -r flag is needed, but if you want to add it, then ok.

The comment about flags is mainly about all the other flags specified in flagfmt - they're not used at all.  E.g. this:
```
flagfmt='A,C,F,H,L,R,U,a,c,d,f,i,h,l,p,r,s,t,u,1'
args='[file ...]'
```

...should be just:
```
flagfmt='r'
```

PR #879, opened 2018-10-30
exec(usbd) args weren't nil terminated, so broke on init

gmacd:
This fixes the the first part of the long-standing USB init bug we've had.
Can now use usb keyboard by adding '-device usb-kbd' to the qemu options, but it seems hyper-sensitive - press a button and get 10 chars.  There also seems to be some warnings about a descriptor on usb init.  That's the next step.  Also usb mouse doesn't seem to work from startup.  It /seems/ to work if you add the mouse (in qemu) mid-session though.  Maybe?

### DIFF
--- a/sys/src/9/boot/boot.c
+++ b/sys/src/9/boot/boot.c
@@ -328,7 +328,7 @@ static void
 usbinit(void)
 {
 	static char usbd[] = "/boot/usbd";
-	static char *argv[] = {"usbd"};
+	static char *argv[] = {"usbd", nil};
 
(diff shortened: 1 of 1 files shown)


elbing:
maybe the multiple chars bug couldn't be only in usb, we haven't standard devcons inherited from plan9, it was... refactored. Since mine is the blame when we discarded out of kernel consoles and the primitive Aki's cons (it hadn't a proper dev/cons in userspace), maybe I screw anything up somewhere...

PR #883, opened 2018-11-10
Adding some CPUs

elbing:


gmacd:
So before this gets merged in, where is the logic to pick the freq documented?  Linux and cpuid think my CPU (Intel Core i7-8550U) should be 1800MHz, but harvey thinks 1666MHz.

gmacd:
Seems good enough for now, but we should look into this in the future:
- is it accurate?
- maybe some obvious indication if a cpu is unrecognised (so we can add quickly)

PR #894, opened 2019-03-06
Re-add telnet and rlogin

gmacd:
Needs segments to be fixed before telnetd will work

This partially reverts commit ccdb6fd39c34ffdaf39ed685fa615c5671ed17e3.


dancrossnyc:
Telnet I kind of get; by why rlogin!? That should be relegated to the scrapheap of history.

gmacd:
No strong opinion on rlogin - I have no use for it tbh :)

elbing:
Dan it's all the same. At the end all of them are fronts of telnetd, even netssh. I need a way of connecting without drawterm for some gcc tasks which will require hours.

PR #895, opened 2019-03-06
Re-add vganvidia - kernel support for old nvidia cards

gmacd:
Re-add from https://github.com/0intro/plan9
Modified to build.  Fails when on when vga connects to nvidiammio segment.  So another reason to fix segments.


elbing:
Outdated.

PR #899, opened 2019-03-09
Add acid to default build, fix lots of build errors

gmacd:
Removed the plan 9 extensions compile flag, fixed all errors caused by this, and add it to the build.

Can run `acid /bin/ls`.  Stops at main, can then continue.  Haven't tried more than that, but it seems like a good start :)


elbing:
update

PR #903, opened 2019-07-20
Fix bug where tmp wasn't bound properly since syscall doesn't exist

gmacd:
If anyone has a better way of doing it (since I'm crap at rc) go ahead

(This was breaking a bunch of go tests, btw(


sevki:
Looks very reasonable to me, not that I am any better at rc

PR #913, opened 2020-02-08
Removing pic and bzip2 now in APEX repo

elbing:


elbing:
bzip2 is ported without ansi requirements.

PR #917, opened 2020-06-30
The empire strikes back

rminnich:
Bring a.out format back. Go is not going to be flexible on this point.

Feel free to tell me how to help make this compile.
e.g. where is l2be again? I forget.

Signed-off-by: Ronald Minnich <rminnich@gmail.com>

### DIFF
--- a/sys/src/9/port/sysproc.c
+++ b/sys/src/9/port/sysproc.c
@@ -380,16 +380,12 @@ execac(Ar0* ar0, int flags, char *ufile, char **argv)
 	/*
 	 * #! has had its chance, now we need a real binary.
 	 */
-	// hook for a.out
-	/* 	magic = l2be(hdr.Exec.magic); */
-	/* if(hdrsz != sizeof(Hdr) || magic != AOUT_MAGIC) */
-	/* 	error(Ebadexec); */
 	// aoutldseg does no file i/o for its test.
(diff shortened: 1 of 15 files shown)


rminnich:
note: doesn't built.

rminnich:
note also: intent is that sysexec will call this when it figures out it is an a.out. Lots of crap deleted since it's already done.

rminnich:
builds.

next steps. Fill out the loading of exec
run a go binary for plan 9 which does nothing
watch it error

start filling in plan 9 system call stubs -- that can be done in syscall(), believe it or not

ratrace will be my friend

If all goes well, plan 9 go runs on harvey shortly.

sys/src/9/port/sysproc.c
```diff
@@ -666,6 +685,8 @@ execac(Ar0* ar0, int flags, char *ufile, char **argv)
 		up->procctl = Proc_toac;
 		up->prepagemem = 1;
 	}
+	DBG("up %p: plan 9 proc\n", up);
```
gmacd, line 688: It's not always a plan9 proc at this point

rminnich: ah right I had condition DBG on plan 9. I think I can drop this print now.

Thanks!

PR #920, opened 2020-07-02
Bringing back Ken cc: first step, libcc as a common lib for compilers…

elbing:
… and utilities

### DIFF
--- a/sys/include/cc.h
+++ b/sys/include/cc.h
@@ -1,0 +1,800 @@
+/*
+ * This file is part of the UCB release of Plan 9. It is subject to the license
+ * terms in the LICENSE file found in the top-level directory of this
+ * distribution and at http://akaros.cs.berkeley.edu/files/Plan9License. No
+ * part of the UCB release of Plan 9, including this file, may be copied,
+ * modified, propagated, or distributed except according to the terms contained
+ * in the LICENSE file.
+ */
+
+#include <u.h>
+#include <libc.h>
+#include <bio.h>
+#include <ctype.h>
+
+// For Harvey
+typedef	unsigned short	ushort;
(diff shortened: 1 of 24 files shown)


go.sum
```diff
@@ -1,5 +1,7 @@
 github.com/Harvey-OS/ninep v0.0.0-20190302184144-8010cab8eac9 h1:vcT+syCMklD8M/k/PcNYehmCRYrp8cyO7pjlZC1/vwg=
 github.com/Harvey-OS/ninep v0.0.0-20190302184144-8010cab8eac9/go.mod h1:YmLTajv/R+cjEsJ8/vfGOckeEBbz06tgHlYBSpbe+eo=
+github.com/Harvey-OS/ninep v0.0.0-20200701230449-5148a8f50134 h1:0Qo7JlCQOUMhO2+L9Mtujhz/cWykNG9Mmx5goPGLCLA=
```
sevki, line 3: weren't we supposed to .gitignore thesse?

sevki:
```
rm go.sum 
git add go.sum
```
`.gitignore`
```
+ go.sum
+ go.mod
```
`go.mod`
```
go 1.13

require (
)

```

go.sum
elbing: go.* removed from PR.

fhs: The `go.mod` and `go.sum` files are generally checked into git. From https://blog.golang.org/using-go-modules:
> The go command uses the go.sum file to ensure that future downloads of these modules retrieve the same bits as the first download, to ensure the modules your project depends on do not change unexpectedly, whether for malicious, accidental, or other reasons. Both go.mod and go.sum should be checked into version control.

PR #931, opened 2020-07-06
bootstrap: add gotip to bootstrap script

sevki:
bootstrap script will now build gotip for plan9 GOOS
and leave it in the util folder, so after boot gotip can download
and run/compile the tip of go tree.

Signed-off-by: sevki <s@sevki.org>


.github/workflows/harvey.yml
```diff
@@ -26,9 +26,15 @@ jobs:
         sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
         sudo add-apt-repository 'deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-10 main' -y
         sudo apt-get update
-        sudo apt-get install -y --force-yes ${{ matrix.cc }} rc golang-1.11-go
+        sudo apt-get install -y --force-yes ${{ matrix.cc }} rc golang-1.11-go qemu
```
rminnich, line 29: could we do 1.14?

util/SERIALQRUN
```diff
@@ -1,10 +1,10 @@
 sudo qemu-system-x86_64 -s -cpu Opteron_G1 -smp 8 -m 2048  \
```
rminnich, line 1: The point of this script was to work on serial console only machines. You've removed the serial argument, left the nographic argument in place, and added via arguments ... this is no longer a serial port invocation. I'm wondering why the change is needed.

sevki:
that's high praise!

PR #934, opened 2020-07-07
Fix build for gcc version 10

fhs:
gcc 10 defaults to `-fno-common`, which results in a lot of "multiple
definition" errors when linking. As suggested by the gcc documentation,
we workaround the issue for now by setting the `-fcommon` flag:
https://gcc.gnu.org/gcc-10/porting_to.html#common


sevki:
shoots from gcc docs
image
and its nothing but net

PR #938, opened 2020-07-14
Change default 9600 baud to 115200

rminnich:
This has been a pain for years, finally fixed.


rminnich:
arguably, we might not want to set stuff at all; bios usually does it nowadays, but this fixes one big mess for the atomic pi.

PR #940, opened 2020-07-15
bootstrap.sh: remove GOBIN for cross-compiled binary

orangecms:
Go otherwise complains:
go get: cannot install cross-compiled binaries when GOBIN is set

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -6,7 +6,6 @@ GO111MODULE=on GOBIN="$(pwd)/util" go get -u github.com/Harvey-OS/ninep/cmd/ufs
 # GO111MODULE=on GOBIN="$(pwd)/util" go get -f -u bldy.build/bldy
 GO111MODULE=on GOBIN="$(pwd)/util" go get ./util/src/harvey/cmd/...
 
-GOOS=plan9 GOARCH=amd64 GO111MODULE=on go get golang.org/dl/gotip
 
 # this will make booting a VM easier
 mkdir -p tmp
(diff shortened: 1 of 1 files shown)


bootstrap.sh
```diff
@@ -6,8 +6,7 @@ GO111MODULE=on GOBIN="$(pwd)/util" go get -u github.com/Harvey-OS/ninep/cmd/ufs
 # GO111MODULE=on GOBIN="$(pwd)/util" go get -f -u bldy.build/bldy
 GO111MODULE=on GOBIN="$(pwd)/util" go get ./util/src/harvey/cmd/...
 
-GOOS=plan9 GOARCH=amd64 GO111MODULE=on GOBIN="$(pwd)/util" go get golang.org/dl/gotip
-
+GOOS=plan9 GOARCH=amd64 GO111MODULE=on go get golang.org/dl/gotip
```
sevki, line 9:
```suggestion
```

orangecms: Oh I see, thanks! I will wrap my head around that. :)

rminnich:
could we, rather, start creating bin directories in form 
util/linux_amd64/bin/...
util/plan9_amd64/bin/...
etc.
is that possible? We need to start thinking in terms of heterogeneity now.

The plan 9 naming won't quite do it, since the only differentiate on architecture.

Daniel, nice to see you here.

bootstrap.sh
fhs: Beside the home directory issue, you'd need to a working `git` command to clone the source and an existing Go distribution already installed to bootstrap Go. Maybe we can write a rc script to bootstrap go tip using the plan9 binaries on this page: http://9legacy.org/download.html

sevki: Hey @orangecms I would like to commit this suggestion and merge this... is that ok?

orangecms:
Sorry, was a bit unsure how to go about this for now.

So the line is just removed for now, and I am creating a new issue for the goal of getting a Go toolchain in place. :)

PR #941, opened 2020-07-16
Fix multiboot args

rminnich:
clean up rsdp parsing. Vendors don't follow ACPI rules, there's a shock.


elbing:
Do you wanna keep v=1 for debug by default?

PR #942, opened 2020-07-21
rename master to main

sevki:
Signed-off-by: sevki <s@sevki.org>

### DIFF
--- a/.github/workflows/harvey.yml
+++ b/.github/workflows/harvey.yml
@@ -2,9 +2,9 @@ name: harvey
 
 on:
   push:
-    branches: [ master ]
+    branches: [ main ]
   pull_request:
-    branches: [ master ]
+    branches: [ main ]
 
 jobs:
   build:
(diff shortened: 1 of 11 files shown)


sys/src/go.mod
```diff
@@ -0,0 +1,3 @@
+module harvey-os.org
```
fhs, line 1: Why do we need this file? The module at this url points to the [go repo](https://github.com/Harvey-OS/go/blob/98bcd60fd161e48f21dae558a19bec6360039962/go.mod#L1):
```
$ curl -L 'http://harvey-os.org/?go-get=1'
<html><head><meta name="go-import" content="harvey-os.org git https://github.com/Harvey-OS/go"></head></html>
```

sevki: accidental commit :/ thanks for the heads up :)

gmacd:
My personal opinion is that these easy changes to the repo branch names are good.

If you also want to change comments, then fine, but I think for a project like harvey (lots of historical code, small team, low importance/visibility) shouldn't fell it necessary to spend time rewriting comments and changing names based on new norms.

PR #945, opened 2020-07-23
Add support for PC Engines APU2

rminnich:
Add the signature.

Fix allocator issues. coreboot reports page 0 as reserved and this breaks
the address space map allocator. Coreboot also reports memory starting at
0xc0000 as usable, and the asm code would toss that -- in the process, tossing
all of memory.

asm has turned out to be a bit fragile. It can't function without having memory
start at 0. But the code is a bit subtle, so be careful if you decide to adjust it.

Add a comment to mmuinit to explain an error message that is seen all too often.


rminnich:
please test on real hardware

PR #949, opened 2020-07-24
more apu2 fixes: dial down spu, and fix a broken compare

rminnich:


rminnich:
Remember I keep saying it's a game of inches? Here's a typical example: == where it should be !=

PR #950, opened 2020-07-24
uarti8250: make COM1 and COM2 poll, or any uart with poll or tbdf==-1

rminnich:
now that UARTs are showing up in #t again, any attempt to use them, e.g.
echo hi >> '#t/eia0'
will get a panic, as our interrupt support code no longer supports
ancient hardware with a tbdf of -1.

Note that, cleverly, uart interrupts are never even enabled until the
device is opened, which is why we did not see this problem earlier

We made this change I believe in the mod 2000s as part of the Fast-OS effort,
as we tried to drag the code forward from its 1990 base. It's a shame we
did not have code control back then.

i8250enable now checks to see if ctlr->tbdf is -1, or ctlr->poll is set;
if so, it disables interrupt enable.

The i8250poll code now runs, finally.

But what broke between 2011 and how such that serial consoles stopped working.
Therein lies a puzzle.

### DIFF
--- a/sys/src/9/386/uarti8250.c
+++ b/sys/src/9/386/uarti8250.c
@@ -480,6 +480,7 @@ i8250interrupt(Ureg* ureg, void* arg)
 	uart = arg;
 
 	ctlr = uart->regs;
+	outb(0x3f8, '.');
 	for(iir = csr8r(ctlr, Iir); !(iir & Ip); iir = csr8r(ctlr, Iir)){
 		switch(iir & IirMASK){
 		case Ims:		/* Ms interrupt */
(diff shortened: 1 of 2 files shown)


rminnich:
btw the second commit is only marginally better. It does, however, work for this trivial case of 2 uarts. And the old code did not.

The uarts are in a pre-defined array and element has a next pointer.  some of them might be working, some not.

consider this array of uarts in a linked list.
a->b->c->d

The for loop walks the list, and its goal is to test which ones work, and if they work, allocate io ports. 
If not, delete the uart from the list by changing next pointers of the previous one. (it does this with the (uart-1) pointer kludge)

There may be 0, 1, or more working uarts. You might return this:
a->b->c->d
or
a->d (and in this case the (uart-1) pointer hackery is wrong

Anyway, in case you think the code is bogus, it is,  *if there are more than 2 uarts*. But there are only two, ever; so I doubt it matters much. I think it's too much effort for 2 uarts. But feel free to change it. Just be careful; it's been broken a lot.

rminnich:
don't merge this, just review, there's a debug print that has to be removed.

rminnich:
ok, feel free to merge if you want. It's not great, it's just better  than what we had.

rminnich:
let's take this in and, if we see a problem down the way, we can undo it.

PR #951, opened 2020-07-25
Move go repo and copy ninep repo into harvey (sys/src/go)

gmacd:
Moves the code for ufs and tmpfs into the harvey repo.  The go repo can now be deleted.  Ninep can remain since others use it, but probably worth putting a notice saying the code has migrated.

todo
- Change the build scripts to use ufs in harvey rather than ninep repo
- Think about how we can manage go binaries for harvey vs for the host


sys/src/go/cmd/tmpfs/tmpfs.go
```diff
@@ -0,0 +1,344 @@
...
+	"os"
+	"sync"
+
+	"harvey-os.org/internal/tmpfs"
+	"harvey-os.org/pkg/ninep"
+	"harvey-os.org/pkg/ninep/protocol"
+)
+
+var (
+	networktype = flag.String("ntype", "tcp4", "Default network type")
```
rminnich, line 18: could it just be "net", that's kind of consistent with the Go nomenclatures
tmps -net=tcp
etc.etc.

sys/src/go/cmd/tmpfs/tmpfs.go
```diff
@@ -0,0 +1,344 @@
...
+	netaddr     = flag.String("addr", ":5641", "Network address")
+	debug       = flag.Int("debug", 0, "Print debug messages")
```
rminnich, line 20: is this a level? If not a level, make it a Bool.

sys/src/go/cmd/tmpfs/tmpfs.go
```diff
@@ -0,0 +1,344 @@
...
+	if version != "9P2000" {
+		return 0, "", fmt.Errorf("%v not supported; only 9P2000", version)
+	}
+	return msize, version, nil
+}
+
+// Rattach attaches a fid to the root for the given user.  aname and afid are not used.
+func (fs *fileServer) Rattach(fid protocol.FID, afid protocol.FID, uname string, aname string) (protocol.QID, error) {
+	if afid != protocol.NOFID {
+		return protocol.QID{}, fmt.Errorf("We don't do auth attach")
```
rminnich, line 65: the linux 9p fs code expects very specific strings, and if a return like this does not match, it returns some useless number. Consider sending a message known to linux 9p client, if possible. If not, that's fine too, but I've been burned by this.

sys/src/go/cmd/tmpfs/tmpfs.go
```diff
@@ -0,0 +1,344 @@
...
+					// first elementwise walk that failed. The value of nwqid cannot be
+					// zero unless nwname is zero. Also, nwqid will always be less than
+					// or equal to nwname. Only if it is equal, however, will newfid be
+					// affected, in which case newfid will represent the file reached
+					// by the final elementwise walk requested in the message.
+					//
+					// to sum up: if any walks have succeeded, you return the QIDS for
+					// one more than the last successful walk
+					if i == 0 {
+						return nil, fmt.Errorf("file does not exist")
```
rminnich, line 119: consider making consts via errors.New for all these strings, as is common. Try to make sure they match what Linux expects. See if any are in the go syscall package for plan 9 already?

sys/src/go/cmd/tmpfs/tmpfs.go
```diff
@@ -0,0 +1,344 @@
...
+	}
+
+	// TODO Check executable
+
+	return f.Qid(), fs.ioUnit, nil
+}
+
+// Rcreate not supported since it's a read-only filesystem
+func (fs *fileServer) Rcreate(fid protocol.FID, name string, perm protocol.Perm, mode protocol.Mode) (protocol.QID, protocol.MaxSize, error) {
+	return protocol.QID{}, 0, fmt.Errorf("filesystem is read-only")
```
rminnich, line 153: same comment for all strings

sys/src/go/cmd/tmpfs/tmpfs.go
```diff
@@ -0,0 +1,344 @@
...
+	}
+
+	// Bind and listen on the socket.
+	listener, err := net.Listen(*networktype, *netaddr)
+	if err != nil {
+		log.Fatalf("Listen failed: %v", err)
+	}
+
+	ufslistener, err := newTmpfs(arch, func(l *protocol.Listener) error {
+		l.Trace = nil // log.Printf
```
rminnich, line 337: well here's an ideal place to use your Debug switch.

sys/src/go/cmd/tmpfs/tmpfs_test.go
```diff
@@ -0,0 +1,235 @@
...
+
+	arch := tmpfs.ReadImage(createTestImage())
+
+	fs, err = newTmpfs(arch)
+	if fs == nil || err != nil {
+		t.Fatal("should not have failed")
+	}
+}
+
+func TestMount(t *testing.T) {
```
rminnich, line 81: TODO: use iostats to trace a session on this, and from that, create a table driven test. 

Someday.

sys/src/go/cmd/tmpfs/tmpfs.go
gmacd: net clashes with the "net" import

gmacd: go pkg uses "network" as the arg name, so could use that (here and for ufs for consistency)

gmacd: or just use net as the flag name and keep the var name as is

gmacd:
I'll get it in and tweak after - this PR was just to move the code into harvey

PR #952, opened 2020-07-25
bootstrap uses ufs from this repo, tmpfs better error strings and debug

gmacd:

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -2,8 +2,7 @@
 
 echo Building the build tool...
 
-GO111MODULE=on GOBIN="$(pwd)/util" go get -u github.com/Harvey-OS/ninep/cmd/ufs
-# GO111MODULE=on GOBIN="$(pwd)/util" go get -f -u bldy.build/bldy
+(cd sys/src/go; GOBIN="$HARVEY/util" go install ./...)
 GO111MODULE=on GOBIN="$(pwd)/util" go get ./util/src/harvey/cmd/...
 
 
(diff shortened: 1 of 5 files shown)


bootstrap.sh
```diff
@@ -2,8 +2,7 @@
 
 echo Building the build tool...
 
-GO111MODULE=on GOBIN="$(pwd)/util" go get -u github.com/Harvey-OS/ninep/cmd/ufs
-# GO111MODULE=on GOBIN="$(pwd)/util" go get -f -u bldy.build/bldy
+(cd sys/src/go; GOBIN="$HARVEY/util" go install ./...)
```
fhs, line 5: Run `go mod tidy` so that it'll remove `github.com/Harvey-OS/ninep` from top-level `go.mod` file?

rminnich: I'm glad you're here ... I don't understand modules, still :-)

sys/src/go/internal/ufs/filesystem.go
```diff
@@ -446,15 +440,16 @@ func (e *FileServer) Rwrite(fid protocol.FID, o protocol.Offset, b []byte) (prot
 	return protocol.Count(n), err
 }
 
-func NewUFS(opts ...protocol.ListenerOpt) (*protocol.Listener, error) {
+func NewUFS(root string, debug int, opts ...protocol.ListenerOpt) (*protocol.Listener, error) {
```
rminnich, line 443: so ... we can do this later but convention would be to make ALL of these opts ...protocol.ListenerOpt
NewUFS(opts ...protocol.ListenerOpt)
and then we would supply things like
WithRoot(root string)
WithDebug()
and so on.
 Up to you if you want to do that now but it would be nice to get it this way from the start.

OTOH if you really prefer it to look this way, that's fine too, since it kind of matches your command.

PR #955, opened 2020-07-31
devramfs: get it to work

rminnich:
devramfs violated a lot of rules about writing drivers, in its not quite
right user (or lack of) of error etc.

It also panicked the kernel.

This verion will let your write up to a 16MiB file.

Needs work, but it's far more workable than what we had.

### DIFF
--- a/sys/src/9/port/devramfs.c
+++ b/sys/src/9/port/devramfs.c
@@ -7,36 +7,37 @@
  * in the LICENSE file.
  */
 
-#include	"u.h"
-#include       "../port/lib.h"
-#include       "mem.h"
-#include       "dat.h"
-#include       "fns.h"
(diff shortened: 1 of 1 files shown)


sys/src/9/port/devramfs.c
```diff
@@ -90,178 +91,204 @@ ramreset(void)
...
 			error(Eperm);
 		} else {
 			prev->sibling = file->sibling;
 		}
 	}
-	if(!(file->perm & DMDIR)){
-		free(file->data);
+	if(!(file->perm & DMDIR)) {
+		for(int i = 0; i < nelem(file->blocks); i++)
+			free(file->blocks[i]);
```
rminnich, line 238: don't forget, nil is a legal value to pass to free.

PR #957, opened 2020-07-31
re-enable the vflag command line parameter; add printallsyscalls

rminnich:
now -v1 will get your verbose, else not.

Adding printallsyscalls=1 will get all system calls printed.

### DIFF
--- a/sys/src/9/amd64/main.c
+++ b/sys/src/9/amd64/main.c
@@ -52,7 +52,7 @@ static int maxcores = 1024;	/* max # of cores given as an argument */
 static int numtcs = 32;		/* initial # of TCs */
 
 char dbgflg[256];
-static int vflag = 1;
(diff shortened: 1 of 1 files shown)


sys/src/9/amd64/main.c
```diff
@@ -161,6 +162,9 @@ options(int argc, char* argv[])
 			if(strcmp(env[0], "numtcs") == 0){
 				numtcs = strtol(env[1], 0, 0);
 			}
+			if(strcmp(env[0], "printallsyscalls") == 0){
```
gmacd, line 165: We should document all these flags somewhere.

sevki: YESSSS!!!! https://github.com/Harvey-OS/harvey/issues/958 follow up issue here!

PR #960, opened 2020-08-02
Restore std detection

rminnich:
An STD is when the Direction flag is inherited by the kernel from user space
and copies run backwards. A symptom is you see the kernel sending R messages
instead of T messages, when the kernel is not serving any 9p sessions.

when we reverted to the NIX scheduler in 42eba8ecadf61a25fb88710d8aae0f5f24017b14
we lost STD detection. We still have protection, but it pays to check, since
the consequences are dire.

Bring it back.


rminnich:
The command "sudo apt-get install -y --force-yes $CC rc" failed and exited with 100 during .

annoying.

PR #962, opened 2020-08-03
tasunlock: add extra coherence (Richard Miller)

rminnich:
http://9legacy.org/9legacy/patch/9-taslock-unlock.diff

from 2016.


rminnich:
This does not fix the Go smp problem with locks. That said, it seems prudent.

elbing:
What about iunlock() a few lines below? is not the same case?
```
l->m = nil;
l->key = 0;
coherence();
```
Maybe Richard didn't see it.

rminnich:
nice catch, and I missed that. Hang on, I'll add it.

rminnich:
Thanks for catching that!

This does not fix the Go program but it's nice to have.

rminnich:
Richard did see it, I just missed it :-)

PR #963, opened 2020-08-04
implement invlpg

rminnich:
It was never turned on because I never understood the syntax.
This does NOT solve the Go problem on SMP.

### DIFF
--- a/sys/src/9/amd64/l64v.S
+++ b/sys/src/9/amd64/l64v.S
@@ -253,10 +253,7 @@ wrmsr:
 
 .global invlpg
 invlpg:
-#	MOVQ	%rdi, va+0(FP)
-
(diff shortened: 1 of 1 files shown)


sys/src/9/amd64/l64v.S
```diff
@@ -253,10 +253,7 @@ wrmsr:
 
 .global invlpg
 invlpg:
-#	MOVQ	%rdi, va+0(FP)
-
-#	INVLPG	va+0(FP)
-
+	invlpg 0(%rdi)
```
dancrossnyc, line 256: Drop the `0`? It shouldn't be necessary.

rminnich: good point. done.

PR #966, opened 2020-08-09
Add initial support for Intel I219 ethernet.

gmacd:
Add initial support for Intel I219 ethernet.
This allows the controller to identify the I219 and makes it usable.
Takes a couple of minutes to startup, but is able to communicate across a network, and can be used to boot up with a remote ufs.

### DIFF
--- a/sys/src/9/amd64/ether82563.c
+++ b/sys/src/9/amd64/ether82563.c
@@ -485,6 +485,7 @@ enum {
 	i210,
 	i217,
 	i218,
+	i219,
 };
 
 static char *tname[] = {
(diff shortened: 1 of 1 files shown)


gmacd:
Slow to start up as in a couple of minutes.  The boot process then continues by reading files from ufs through the network controller - so I've verified it works to that extent.  I can make the commit message clearer at some point.

rminnich:
please make it clearer now. It's really hard to go back and fix a commit message. Pretty please?

gmacd:
Should be clearer now.

gmacd:
Need another approval after the rebase

PR #967, opened 2020-08-09
Add support for Intel Z170 chipset to sdiahci.c

gmacd:
Seems to have difficulty with one of the ports, causing it to hang.  Not sure why at this point, but if you skip it, it boots.

### DIFF
--- a/sys/src/9/amd64/sdiahci.c
+++ b/sys/src/9/amd64/sdiahci.c
@@ -2002,8 +2002,10 @@ didtype(Pcidev *p)
 		    (p->did & 0xfffe) == 0x2922 ||	/* ich9 */
 		    p->did == 0x3a02 ||			/* 82801jd/do */
 		    (p->did & 0xfefe) == 0x3a22 ||	/* ich10, pch */
-		    (p->did & 0xfff8) == 0x3b28)	/* pchm */
+		    (p->did & 0xfff8) == 0x3b28 ||	/* pchm */
+		    p->did == 0xa102) {			/* q170/q150/b150/h170/h110/z170/cm236 */
 			return Tich;
+		}
 		break;
 	case Vatiamd:
 		if(p->did == 0x4380 || p->did == 0x4390 || p->did == 0x4391){


sys/src/9/amd64/sdiahci.c
```diff
@@ -2002,7 +2002,8 @@ didtype(Pcidev *p)
 		    (p->did & 0xfffe) == 0x2922 ||	/* ich9 */
 		    p->did == 0x3a02 ||			/* 82801jd/do */
 		    (p->did & 0xfefe) == 0x3a22 ||	/* ich10, pch */
-		    (p->did & 0xfff8) == 0x3b28)	/* pchm */
+		    (p->did & 0xfff8) == 0x3b28 ||	/* pchm */
+		    p->did == 0xa102)			/* q170/q150/b150/h170/h110/z170/cm236 */
```
sevki, line 2006:
```suggestion
		    p->did == 0xa102){ /* q170/q150/b150/h170/h110/z170/cm236 */
return Tich;
}

**Comment by rminnich** [comment]
The easiest thing to do is another commit with clang-format. I think we should do that with the whole code base, personally.

**Comment by gmacd** [comment]
It hangs in ahciidle for port 5.  This happens during boot, when calling iapnp in sdiahci (which then calls newctlr, then achiidle).  It tries to dereference port->cmd as so: ```if((*p & Arun) == 0)```

**Comment by gmacd** [comment]
There shouldn't be anything to unusual about the hardware, other than it's more recent than the code.

File: sys/src/9/amd64/sdiahci.c (hunk 1 above)
Comment (rminnich) [reply sys/src/9/amd64/sdiahci.c:2006]: This kind of thing should be fixed with clang-format maybe?

**Comment by sevki** [comment]
@rminnich we have [.clang-format](https://github.com/Harvey-OS/harvey/blob/main/.clang-format) at the root of the repo but it breaks the coding convention.
the specific issue was pointer alignment. consider the trivial example here
```c
Exectab*
lookup(Rune *r, int n) {
  return 0;
}
```
clang format unfortunately will format the pointer either to left
```c
Exectab*
lookup(Rune* r, int n) {
  return 0;
}
```
right:
```c
Exectab *
lookup(Rune *r, int n) {
  return 0;
}
```
or everyone's favorite middle :shrug:
```c
Exectab *
lookup(Rune *r, int n) {
  return 0;
}
```
Last time I tried to make clang-format a required check, that's where it fell apart. 

And it wasn't even a stylistic issue, it just broke semantics. @dancrossnyc explained why left aligned pointers [are evil](https://github.com/Harvey-OS/harvey/pull/255#issuecomment-239623972)

Having said that, while it might break the semantics, it enforces style consistency. Which we're not doing great on :/
For instance
right aligned: https://github.com/Harvey-OS/harvey/blob/main/sys/src/9/amd64/apic.c#L163
left aligned: https://github.com/Harvey-OS/harvey/blob/main/sys/src/9/amd64/apic.c#L192 
so it ended up that the :smiling_imp: left aligned pointers made their way in anyways. 

Anyways maybe we should take a [page](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=8m43s) from the go book, pick a style and live with it.  We can start by enabling a required check [only for the lines that have been edited. ](https://github.com/llvm-mirror/clang/blob/master/tools/clang-format/clang-format-diff.py) What do you think?

gmacd:
Tidied code and comment.  As mentioned, I think the AHCI port 5 issue is a separate issue to the identification of the motherboard, so can be handled separately.

rminnich:
take a page from the go book, pick a style, and live with it. Accept it will never have perfect results. Accept that.

Let's just do it.

PR #969, opened 2020-08-18
Enable u-root

rminnich:
Once the changes are committed for the tmpfs command,
we can start to use u-root.

These changes enable u-root usage for harvey.


lib/namespace
```diff
@@ -24,6 +24,11 @@ bind -a '#r' /dev
 mount -a /srv/slashn /n
 mount -a /boot /bin
 
+# uroot
+tmpfs /bin/uroot &
```
sevki, line 28: is tmpfs a directive the namespace file knows? 
or should this be `import tmpfs`

sys/src/9/amd64/build.json
```diff
@@ -1,8 +1,16 @@
 [
 	{
 		"Name": "cpu",
+		"Pre": [
+		    "u-root -o uroot.cpio plan9",
```
sevki, line 5: we need u-root in `bootstrap.sh`

sevki (changes requested):

lib/namespace
rminnich: tmpfs is the program tmpfs

sys/src/9/amd64/build.json
rminnich: not sure what you mean.

rminnich: OH! now I know what you mean! 

OK I'll do that first, this PR is not quite right yet anyway, thanks!

rminnich:
ok this one works now.

gmacd (approved):

gmacd (approved):

rminnich:
well, SHIT, bootstrap and build are separate
mv $WORK/b001/exe/a.out /home/runner/work/harvey/harvey/linux_amd64/bin/u-root
1900
rm -r $WORK/b001/
ok

orangecms:
#975 is now experimental to see where things are being put. It looks like the Go modules thing and the build tools are not yet cooperating as we'd like. My current approach is to `cp -r` the `$GOPATH/mod/` things to `$GOPATH/src/`. It is ugly, but at least I could get u-root to build a cpio. Now the question is how it would work properly. I will have a look at the Golang docs, hope it'll reveal more details. :)

Issue #319, opened 2016-10-07
Make our builds reproducible


rminnich:
https://reproducible-builds.org/

gmacd:
Spring cleaning - closing old issues

PR #970, opened 2020-08-19
Add u-root and tmpfs to the bootstrap

rminnich:
This ensures that we have u-root and tmpfs for inclusion in
kernels.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -1,10 +1,22 @@
 #!/bin/sh
 
+set -e
+
+export GOBIN=$(pwd)/$(go env GOHOSTOS)_$(go env GOHOSTARCH)/bin
+echo GOBIN is now $GOBIN
+
 echo Building the build tool...
+GO111MODULE=on go get ./util/src/harvey/cmd/...
 
-GOBIN="$(pwd)/util" go get -u harvey-os.org/cmd/ufs
-GO111MODULE=on GOBIN="$(pwd)/util" go get ./util/src/harvey/cmd/...
+echo Fetching u-root and building it...
+GO111MODULE=on go get github.com/u-root/u-root
 
+echo Fetch harvey-os.org commands and build them into $GOBIN
+GO111MODULE=on go get harvey-os.org/cmd/...
+
+echo FIXME -- once we get more architectures, this needs to be done in sys/src/cmds/build.json
+echo Build tmpfs command into amd64 plan 9 bin
+GO111MODULE=on GOOS=plan9 GOARCH=amd64 go build -o amd64/bin/tmpfs harvey-os.org/cmd/tmpfs
 
 # this will make booting a VM easier
 mkdir -p tmp
@@ -15,7 +27,7 @@ export ARCH=amd64
 # You also need to export your C compiler flavor (gcc, clang, gcc-7...)
 export CC=gcc
 # And build:
-./util/build
+$GOBIN/build
 # See \`build -h' for more information on the build tool.
 
 To enable access to files, create a harvey and none user:
@@ -23,4 +35,8 @@ sudo useradd harvey
 sudo useradd none
 
 none is only required for drawterm/cpu access
+
+Also:
+export HARVEY=$(pwd)
+add $GOBIN to your path
 EOF


gmacd (approved):

bootstrap.sh
```diff
@@ -1,10 +1,19 @@
...
+set -e
+
 echo Building the build tool...
 
 GOBIN="$(pwd)/util" go get -u harvey-os.org/cmd/ufs
 GO111MODULE=on GOBIN="$(pwd)/util" go get ./util/src/harvey/cmd/...
 
+echo Fetching u-root and building it...
+GO111MODULE=off go get github.com/u-root/u-root
+GO111MODULE=off go install github.com/u-root/u-root
```
orangecms, line 12: Oh I will need to keep that in mind, might overwrite other experiments... that installs to `$GOPATH/bin`, so usually `$HOME/go/bin`, right? Could we keep it local here?

sevki, line 12: without the -d flag go get installs already. So we can collapse the two lines in to one
```suggestion
GOBIN="$(pwd)/util"  GO111MODULE=off go get github.com/u-root/u-root
```
I agree with @orangecms we should stick to keeping these local.

bootstrap.sh
```diff
@@ -1,10 +1,19 @@
...
+
+echo Fetch tmpfs and building it into amd64/bin
+GO111MODULE=off go get harvey-os.org/cmd/...
```
sevki, line 15: you don't need this line Harvey-OS/go is being fetched and updated on line 7

sevki (changes requested):

bootstrap.sh
rminnich: Shall we stick with util/ then? 

The one good thing about util/, as opposed to bin/ at the top level, is that nobody will think util/ has harvey binaries in it. 

OTOH ... what about a top level directory called $GOOS_$GOARCH, with a bin/ in it, so we can correctly name binaries?

orangecms: I like that composition, it's similar to what u-root uses for naming cpio files. 👍

rminnich:
why is llvm.org such a clusterfuck? 

Should we just start our own GitHub repo with the build artifacts in it?

sevki:
@rminnich can you link the error messsage? I think when you force pushed, that log went away. or atleast I can't find it

bootstrap.sh
rminnich: oops, I thumbed up my own PR

bootstrap.sh
```diff
@@ -15,12 +27,16 @@ export ARCH=amd64
...
 
 To enable access to files, create a harvey and none user:
 sudo useradd harvey
 sudo useradd none
 
 none is only required for drawterm/cpu access
+
+Also:
+export HARVEY=$(pwd)
+add $GOBINn to your path
```
gmacd, line 41: rogue 'n'

orangecms (approved):
👌

gmacd (approved):

bootstrap.sh
```diff
@@ -1,10 +1,22 @@
...
+export GOBIN=$(pwd)/$(go env GOHOSTOS)_$(go env GOHOSTARCH)/bin
+echo GOBIN is now $GOBIN
+
 echo Building the build tool...
+GO111MODULE=on go get ./util/src/harvey/cmd/...
 
-GOBIN="$(pwd)/util" go get -u harvey-os.org/cmd/ufs
-GO111MODULE=on GOBIN="$(pwd)/util" go get ./util/src/harvey/cmd/...
+echo Fetching u-root and building it...
+GO111MODULE=off go get github.com/u-root/u-root
```
sevki, line 12: why are we turning off modules? `GO111MODULE` doesn't require the target package to have a `.mod` file. 
If it's all the same to you I think we should use modules and pin specific versions of external packages.

bootstrap.sh
```diff
@@ -1,10 +1,22 @@
 #!/bin/sh
 
+set -e
+
+export GOBIN=$(pwd)/$(go env GOHOSTOS)_$(go env GOHOSTARCH)/bin
```
sevki, line 5: I don't think we should be setting globals that we haven't declared. this is going to confuse the hell out of people when their binaries start showing up under `harvey/linux_amd64/bin`, even after they stop playing with harvey. 

IMHO `HARVEYBIN` would be better and wouldn't have as many side effects.

bootstrap.sh
```diff
@@ -15,12 +27,16 @@ export ARCH=amd64
...
 
 To enable access to files, create a harvey and none user:
 sudo useradd harvey
 sudo useradd none
 
 none is only required for drawterm/cpu access
+
+Also:
+export HARVEY=$(pwd)
+add $GOBIN to your path
```
sevki, line 41: if you add `$GOBIN` to your path, why are you calling `$GOBIN/build` on line 30?

sevki (changes requested):

bootstrap.sh
fhs: Also note that Harvey's `go.mod` can influence what version of external dependency is used. Which is why I suggest running `go get` inside a temporary directory that's not part of a module. See https://github.com/golang/go/issues/40276

sevki: But in this case that is precisely what we want, we always want a known and tested version of external packages to be downloaded and installed. I feel this is important since we don't have testing of u-root packages on a plan9 system. It's very likely that someone might break a u-root package and not even realize it's broken for plan9.

fhs: Sorry, I meant dependency of u-root can be influence by harvey's dependency when doing the minimum version selection. For example, u-root uses an older version `golang.org/x/tools` package when installed from the harvey module:
```
$ export GO111MODULE=on
$ cd $HARVEY
$ GOBIN=$(pwd)/util/ go get github.com/u-root/u-root@latest
...
$ go version -m util/u-root 
util/u-root: go1.15
	path	github.com/u-root/u-root
	mod	github.com/u-root/u-root	v6.0.0+incompatible	h1:YqPGmRoRyYmeg17KIWFRSyVq6LX5T6GSzawyA6wG6EE=
	dep	github.com/dustin/go-humanize	v1.0.0	h1:VSnTsYCnlFHaM2/igO1h6X3HA71jcobQuxemgkq4zYo=
	dep	golang.org/x/sys	v0.0.0-20190412213103-97732733099d	h1:+R4KGOnez64A81RvjARKc4UT5/tI9ujCIVX+P5KiHuI=
	dep	golang.org/x/tools	v0.0.0-20190524140312-2c0ae7006135	h1:5Beo0mZN8dRzgrMMkDp0jc8YXQKx9DiJ2k1dkvGsn5A=


$ mkdir /tmp/foo
$ cd /tmp/foo/
$ GOBIN=$(pwd) go get github.com/u-root/u-root@latest
...
$ go version -m u-root 
u-root: go1.15
	path	github.com/u-root/u-root
	mod	github.com/u-root/u-root	v6.0.0+incompatible	h1:YqPGmRoRyYmeg17KIWFRSyVq6LX5T6GSzawyA6wG6EE=
	dep	github.com/dustin/go-humanize	v1.0.0	h1:VSnTsYCnlFHaM2/igO1h6X3HA71jcobQuxemgkq4zYo=
	dep	golang.org/x/mod	v0.3.0	h1:RM4zey1++hCTbCVQfnWeKs9/IEsaBLA8vTkd0WVtmH4=
	dep	golang.org/x/sys	v0.0.0-20200819171115-d785dc25833f	h1:KJuwZVtZBVzDmEDtB2zro9CXkD9O0dpCv4o2LHbQIAw=
	dep	golang.org/x/tools	v0.0.0-20200820010801-b793a1359eac	h1:DugppSxw0LSF8lcjaODPJZoDzq0ElTGskTst3ZaBkHI=
	dep	golang.org/x/xerrors	v0.0.0-20191204190536-9bdfabe68543	h1:E7g+9GITq07hpfrRu66IVDexMakfv52eLZ2CXBWiKr4=
```

sevki: I don't think that's a problem for us. 
One of our goals was (is #319?) to make harvey builds reproducible. 
I feel like using `@latest` in the bootstrap process would cause us to back slide in trying to achieve that goal.  As you've demonstrated it's easy to verify what ever it is we're checking in is correct, if you don't think it's correct, update the mod file so it does the correct thing. It requires a bit more diligence on our part but I'd rather have that than `go get @latest` and have builds randomly break because some external dependency updated.

rminnich: u-root and modules still don't play well. If you've tested this with modules turned on and it works, I'm interested to hear it.

bootstrap.sh
rminnich: it's set *in this script*, for this script and its children. That's not global, right?

bootstrap.sh
rminnich: because I did not add GOBIN to the path, the script just recommends it?

bootstrap.sh
rminnich: GOBIN is for use by the go command. 

I'm a bit lost on the problem of setting a variable in a script that's inherited by its children. This is not a global. export is for a process and its children. I don't see the issue here.

bootstrap.sh
rminnich: did you try it first? I'd like to know that it works. Would be appreciated if you could try a thing before advocating changing that thing.

bootstrap.sh
orangecms: yea it's just a long cat-eof-style output comment - confused me at first as well when not looking closely 👀 

starting here https://github.com/Harvey-OS/harvey/pull/970/files#diff-3af436e6760a0c5f1e3d16a3384853a8L12
which GitHub cuts off (you need to click the expander for the full context;  can be a bit annoying here sometimes, and helpful in other cases)

bootstrap.sh
orangecms: Do you mean that it might cause confusion at some point later on when extending the script @sevki? I.e., global from the script's perspective and all its child processes inheriting it?

sevki: No, I screwed up. I was reviewing this on my phone and saw the entire change as a single block. I missed the folded lines.

bootstrap.sh
sevki: yeah, I did.

case in point with with the modules turned off I get
```
GOBIN is now /home/sevki/src/test/harvey/linux_amd64/bin
Building the build tool...
Fetching u-root and building it...
Fetch harvey-os.org commands and build them into /home/sevki/src/test/harvey/linux_amd64/bin
# harvey-os.org/internal/ufs
../../../go/src/harvey-os.org/internal/ufs/filesystem.go:443:45: undefined: protocol.ListenerOpt
../../../go/src/harvey-os.org/internal/ufs/filesystem.go:443:69: undefined: protocol.Listener
../../../go/src/harvey-os.org/internal/ufs/filesystem.go:458:12: undefined: protocol.NewListener
```
because my `$GOPATH` had `511a25aa7f1542a74d755af15e02503a1d87c71a` checked out which was broken and turning on go modules for all fixed the bootstrap script.

rminnich: thanks, good to know it worked for you, I just pushed it with modules on

bootstrap.sh
sevki: yup. I was reviewing it using the mobile app missed the folded lines. my bad folks. I apologize

rminnich: those folded things can be hell. 

Thanks for the reviews everyone, I did set modules on, what else needs to change?

sevki (approved):
LGTM. CI is happy as well. Let's get this in!

PR #971, opened 2020-08-23
add u-root to initram; show changes for go.mod and go.sum

rminnich:


gmacd:
This matches what I get locally, though I suspect we'd need to go get specific version in bootstrap if we want the files to always be consistent.

gmacd (approved):

orangecms:
I just ran the bootstrap.sh script and got the exact same diff.

orangecms (approved):

fhs:
> This matches what I get locally, though I suspect we'd need to go get specific version in bootstrap if we want the files to always be consistent.

I agree. All the `go get foo` needs to be changed to `go get foo@version`. I'd also add a check in the CI that `go.mod` doesn't change after build.

rminnich:
u-root has not been great about versioning -- can we do with a ref instead?

orangecms:
With the second commit came another `go build`, is that what changed go.sum again?

From https://dev.to/maelvls/why-is-go111module-everywhere-and-everything-about-go-modules-24k#where-are-the-sources-of-the-dependencies-with-go-modules:

> When using Go Modules, the packages that are used during `go build` are stored in `$GOPATH/pkg/mod`.

rminnich:
btw for "reasons", you can't build u-root with modules yet. I'm going to leave it to one of you to fix this in a good way this week -- but you'll need to keep that problem in mind. Should our GOPATH be $HARVEY? 

thanks

PR #973, opened 2020-08-27
Avoid running boot process twice if eia0 not found

gmacd:

### DIFF
--- a/sys/src/9/boot/boot.c
+++ b/sys/src/9/boot/boot.c
@@ -298,15 +298,15 @@ static void catstuff(void)
 	ifd = open("#t/eia0", ORDWR);
 	if (ifd < 0) {
 		print("can't open eia0!!!\n");
+		exits("can't open eia0");
 		return;
 	}
 	print("open: ifd %d\n", ifd);
 	while (1) {
 		sleep(100000);
 	}
-
-	print("all done!\n");
 }
+
 /*
  *  ask user from whence cometh the root file system
  */


sys/src/9/boot/boot.c
```diff
@@ -298,14 +300,16 @@ static void catstuff(void)
 	ifd = open("#t/eia0", ORDWR);
 	if (ifd < 0) {
 		print("can't open eia0!!!\n");
-		return;
-	}
-	print("open: ifd %d\n", ifd);
-	while (1) {
-		sleep(100000);
-	}
+	} else {
```
rminnich, line 303: why not replace the return with an 
exits("can not open eia0");

that would be a one line change

Very nice catch by the way :-)

gmacd: could do, though we should also get rid of the print at the end of the function, since it'll never be reached.

gmacd: done

sys/src/9/boot/boot.c
```diff
@@ -298,15 +298,15 @@ static void catstuff(void)
...
+		exits("can't open eia0");
 		return;
 	}
 	print("open: ifd %d\n", ifd);
 	while (1) {
 		sleep(100000);
 	}
-
-	print("all done!\n");
 }
```
rminnich, line 308: was there a reason to remove the all done print? just wondering.

gmacd: It'll never be reached, so why keep it?

gmacd: What I really mean is - if this could be reached, we should have an exits() call after it.  If it can't be reached, it should be a comment at best (IMO).

gmacd: Anyway, thanks for the review.

PR #974, opened 2020-08-27
remove Travis CI setup

orangecms:
Signed-off-by: Daniel Maslowski <info@orangecms.org>

PR #975, opened 2020-08-27
u-root or not u-root...

orangecms:


sevki (approved):
Let's make sure we merge #974 before this, so Travis removal is seperate from go toolchain update.

PR #979, opened 2020-09-01
Extend initial map to 64M

rminnich:
This will allow us to include a big u-root and that u-root can
contain u-root and ALL harvey binaries, i.e. all of amd64/bin.

We're going to want a better compression than just gzip.

-rw-r--r-- 1 rminnich rminnich 70192397 Aug 31 20:43 /tmp/initramfs.plan9_amd64.cpio.gz
-rw-r--r-- 1 rminnich rminnich 23180974 Aug 31 20:25 /tmp/initramfs.plan9_amd64.cpio.lzma

But we are, definitely, going to be able to have pretty much anything we need ... in the initramfs.
This is going to be pretty slick. Boot harvey, get the full graphical environment, THEN mount the
server from 9p. This would include rio etc. This is gonna be very nice.

Graham, tmpfs needs lzma added.

### DIFF
--- a/sys/src/9/amd64/entry.S
+++ b/sys/src/9/amd64/entry.S
@@ -264,8 +264,8 @@ _warp64:
 	addl	$PGLSZ(1), %edx
 	movl	%edx, PDO(KZERO+3*PGLSZ(1))(%eax)	/* PDE for KZERO 6-8MiB */
 
-	// and up through 12 (and on to 16). This sucks, we'll make it better later. //
-	// We'll just have init the pml2 at compile time. Apologies.
+	// and up through 16 (and on to 64). This sucks, we'll make it better later. //
+	// Note that at some point, we create the whole map.
 	addl	$PGLSZ(1), %edx
 	movl	%edx, PDO(KZERO+4*PGLSZ(1))(%eax)	/* PDE for KZERO 8-10MiB */
 	addl	$PGLSZ(1), %edx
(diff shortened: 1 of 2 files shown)


rminnich:
I did try to write a loop, but ... I'm sick of assembly ... ha ha

rminnich:
well it's not 64, and I hate what I did here, but there you are.

sys/src/9/amd64/entry.S
```diff
@@ -275,6 +275,39 @@ _warp64:
 	addl	$PGLSZ(1), %edx
 	movl	%edx, PDO(KZERO+7*PGLSZ(1))(%eax)	/* PDE for KZERO 14-16MiB */
 
+	// and up to 64! I tried a loop but it's harder than I thought.
```
gmacd, line 278: Worth correcting the comment for the size (and the PR title)

sys/src/9/amd64/entry.S
```diff
@@ -275,6 +275,77 @@ _warp64:
...
+	addl	$PGLSZ(1), %edx
+	movl	%edx, PDO(KZERO+28*PGLSZ(1))(%eax)
+	addl	$PGLSZ(1), %edx
+	movl	%edx, PDO(KZERO+29*PGLSZ(1))(%eax)
+	addl	$PGLSZ(1), %edx
+	movl	%edx, PDO(KZERO+30*PGLSZ(1))(%eax)
+	addl	$PGLSZ(1), %edx
+	movl	%edx, PDO(KZERO+31*PGLSZ(1))(%eax)
+
+	movl	%edx, PDO(KZERO+16*PGLSZ(1))(%eax)
```
gmacd, line 333: Shouldn't this be +32?  Or am I missing something subtle?

rminnich: did I screw this up again? I'll look

rminnich: I had one job ... 
check again ....
it does boot. I expect this is not even needed, I am realizing it may only matter until mmuinit(), but I'll test that later.

gmacd (approved):

PR #981, opened 2020-09-01
Fix a stupid error that broke large images

rminnich:
We can now boot with really big ramfs.


orangecms (approved):
That was, indeed, a sjmple thing to fix; I should've had a closer look, too. :)

Issue #977, opened 2020-08-30
setbanked loop when booting on Thinkpad T440p


fhs:
It gets stuck printing this:
```
setbanked: ra 0x1800 rb 0x0 sz 2 value 0x1
```

Video of the boot here: https://youtu.be/CgPTOjpBCgc

rminnich:
the set banked is not the problem. The problem is it does not have a network. 

Is there anyway you can boot with the rc option, not the network option, so we can poke around a bit. But the network is your issue.

fhs:
I can boot with the rc option, but I don't know how to poke around with all the basic commands (cat, ls, etc.) missing.

I have an i217:
```
0.25.0: net  02.00.00 8086/153a   6 0:f0600000 131072 1:f063f000 4096 2:00003081 32
        Intel Corporation Ethernet Connection I217-LM
```
which is supported by the same driver as i219, so I think I'll give PR #980 a try later today.

gmacd:
/boot/ls /boot/cat etc😁

orangecms:
I have a patch for the verbose spewing; see #982

orangecms:
Could you also drop some `lspci` logs on #984 please? :)

orangecms:
The numerous `setbanked` logs are now fixed with #982 and the network issue is tracked in #984, so we can close this, right?

fhs:
> The numerous `setbanked` logs are now fixed with #982 and the network issue is tracked in #984, so we can close this, right?

Sounds good.

PR #982, opened 2020-09-02
Fix verbosity flag

orangecms:
I have no clue why it was as it was, but it looked odd.

### DIFF
--- a/sys/src/9/amd64/devacpi.c
+++ b/sys/src/9/amd64/devacpi.c
@@ -1751,30 +1751,30 @@ static unsigned int getbanked(uintptr_t ra, uintptr_t rb, int sz)
 	return r;
 }
 
-static unsigned int setbanked(uintptr_t ra, uintptr_t rb, int sz, int v)
+static unsigned int setbanked(uintptr_t ra, uintptr_t rb, int sz, int val)
 {
 	unsigned int r;
-	if(v)print("setbanked: ra %#x rb %#x sz %d value %#x\n", ra, rb, sz, v);
+	if(v)print("setbanked: ra %#x rb %#x sz %d value %#x\n", ra, rb, sz, val);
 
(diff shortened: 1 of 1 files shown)


sys/src/9/amd64/devacpi.c
```diff
@@ -1860,7 +1860,7 @@ static void acpiintr(Ureg *u, void *v)
 	//if (sts & 1)
 	//	print("acpiintr: power button\n");
 	// XXX serve other interrupts here.
-	setpm1sts(sts);
+	setpm1sts(v && 1 || 0);
```
rminnich, line 1863: This is ... cryptic. I don't understand what it is doing?

orangecms: `setpm1sts` is expecting an int, see https://github.com/Harvey-OS/harvey/blob/main/sys/src/9/amd64/devacpi.c#L1791
That is being passed on to `setbanked`, see https://github.com/Harvey-OS/harvey/blob/main/sys/src/9/amd64/devacpi.c#L1754
and that determines whether to print the verbose `"setbanked: ...` messages (https://github.com/Harvey-OS/harvey/blob/main/sys/src/9/amd64/devacpi.c#L1757) which are flooding the console (#977).

orangecms: Um, I guess, since `v` is used in a different sense in `setbanked`, we should give some more verbose (pun intended) var names; this patch here is actually causing another issue. Woops.

orangecms: Different patch coming in :sweat_smile:

rminnich (approved):
now that was a nice catch. And that sucky code -- my fault.

rminnich (approved):
I wonder why we did not get an alias warning. We need to enable that.

PR #985, opened 2020-09-02
Build in u-root support

rminnich:
This change builds in u-root support. u-root is no longer optional.

Images will now be larger, but you will get a lot of benefit.

The build now includes ALL of amd64/bin -- ed, sam, rio, acme, -- all of it.
You won't need to mount anything to bring up graphics. You can then have a better
environment for debugging problems.

The current directories brought in via the -files switch to u-root are:
-files /amd64/bin:amd64/bin -files /rc/bin:rc/bin -files /lib/ndb:lib/ndb

You can add extra elements to the initramfs with the UROOTEXTRAFILES
and UROOTEXTRAPACKAGES. In this next example, we wish to add /sys/include, and include
an awk package written in Go:

UROOTEXTRAPACKAGES=github.com/benhoyt/goawk UROOTEXTRAFILES='-files=/sys/include:sys/include'  CC=gcc ARCH=amd64 ../../../../linux_amd64/bin/build

A note on the use of -files:the -files switch takes an input path and an OPTIONAL output path.
The output path should NOT begin with /. The build tool will,
as always, take absolute paths and prepend them with the harvey path, e.g., /amd64/bin becomes
/home/rminnich/harvey/amd64/bin. This is an original build tool property which provides much convenience.

So in the above example, we decided to add /sys/include from harvey, and we place
it in the repo at sys/include -- it still be at /sys/include in the repo, since the destination
is implicitly rooted.

If we leave off the optional : specifier, e.g. we have
-files=/
it will place the entire harvey tree with the path of harvey, i.e.
-files=/
will place /home/rminnich/harvey into the repo with as /home/rminnich/harvey -- unlikely to be what you want.

This patch does not automatically mount the embedded cpio. That comes in a later patch.

To mount it:
/boot/tmpfs /boot/uroot&
/boot/mount /srv/tmpfs /initramfs
/boot/bind -b /initramfs/amd64/bin /bin

ls

And to get the Go programs:
/boot/bind -b /initramfs/bbin /bin

now you have it all.

Note the first step, before /srv/tmpfs appears, can be REALLY SLOW. It took a full 100
seconds to uncompress on my PC Engines APU 2. We need a progress bar I suppose.


floren:
Looks like it wants the u-root binary in your path to work; it'd be nice if we could build that automatically.

orangecms:
> Looks like it wants the u-root binary in your path to work; it'd be nice if we could build that automatically.

Yup, that just came in through the other PR; now we just need to wait for the rebuild - CI is magic :sparkles:  :)

orangecms (dismissed):

orangecms (approved):

rminnich:
btw, this needs testing. That Go problem with deadlock is not necessarily fixed. Just saw it again on a Thinkpad.

rminnich:
but we need to make this work so .... good times.
I  hope some of you can test on hardware.

orangecms:
Can we check that in QEMU?

rminnich:
ok, time to pull the trigger. In any event, git revert is your friend.

PR #986, opened 2020-09-03
Fix build when u-root and tmpfs sources are absent from $GOPATH

fhs:
u-root expects the sources are there in $GOPATH, so download it before
running it.

Remove an unnecessary tmpfs build since it's already being built during
bootstrap.


gmacd (approved):
Seems fine for now. I wonder if we should change the way it works in the future though.

Previously bootstrap was used to build some build tools that rarely changed. Now it's building things we probably want to update more frequently.

Maybe some of this should go into build.json, if it's fast enough.

fhs:
We can use a temporary `$GOPATH` when running u-root. That way we can also lock in the version of u-root tools. I'd rather see u-root add support for modules though.

PR #987, opened 2020-09-03
WIP: require u-root support

rminnich:
remove a bunch of binaries that are in the compressed initramfs.

For now, to boot, you have to
/boot/tmpfs /boot/uroot&
(wait ...)
/boot/mount /srv/tmpfs /initramfs
/boot/bind /initramfs/amd64/bin /bin

### DIFF
--- a/sys/src/9/amd64/build.json
+++ b/sys/src/9/amd64/build.json
@@ -1,11 +1,6 @@
 [
 	{
 		"Name": "cpu",
-		"Pre": [
-		    "u-root -o uroot.cpio plan9",
-		    "gzip -f -k uroot.cpio",
-		    "go build -o tmpfs harvey-os.org/cmd/tmpfs/"
-		],
 		"Env": [
 			"CONF=cpu",
 			"GO111MODULE=off",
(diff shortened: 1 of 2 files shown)


rminnich:
u-root date is noticeably slower than date. fyi.

rminnich:
yeah we got a problem. There is something going very badly in Go with this on a Thinkpad. We're going to need to do some more testing. Absent things getting better we can't depend on tmpfs as we can't depend on go :-(

sys/src/9/amd64/build.json
```diff
@@ -99,34 +94,11 @@
...
-				"echo": "/$ARCH/bin/echo",
-				"ed": "/$ARCH/bin/ed",
-				"factotum": "/$ARCH/bin/auth/factotum",
-				"fdisk": "/$ARCH/bin/disk/fdisk",
-				"fossil": "/$ARCH/bin/fossil/fossil",
-				"grep": "/$ARCH/bin/grep",
-				"ipconfig": "/$ARCH/bin/ip/ipconfig",
-				"irq": "/$ARCH/bin/acpi/irq",
-				"ls": "/$ARCH/bin/ls",
+				"bind": "/$ARCH/bin/bind",
```
gmacd, line 97: got 2 copies of bind

gmacd: don't we need ipconfig?

gmacd: at least it seems I do!

rminnich:
Until KZERO is 0xffffffff_80000000, this approach work. We'll do it another time.

PR #990, opened 2020-09-07
add potentially hacky switch which makes apu2 ethernet work

floren:


rminnich (approved):
I'd feel better knowing why this worked but ... intel ethernet ... coreboot .. who knows.

PR #991, opened 2020-09-07
Add a /lib to root on boot; grow Nrootfiles

rminnich:
we're adding more files, so we might as well grow it a bit.


rminnich:
John, why did you think we need this again? 
If we just mount -b /initramfs / we'll get the /lib for free.
just checking.

floren:
You're right, it's not needed.

PR #993, opened 2020-09-08
mount: allow specification of mount device, small cleanup

rminnich:
We added the ability to use mount devices other than 'M' 5 years ago.
Somehow this command never reflected it or it was lost in the
great source cleanup a few years back.

If the user specifies -M with an option, e.g. 9, that will be
used as the mount device. A -M with 9 specifies 9p2000.L; it's been
in there for 4 years. It has issues but might give us a path to
easily using ipfs and hugelgupf/p9 without changing those systems.
It is claimed to work with the qemu 9p2000.L server.

Also, do a minor cleanup. The name 'aname' was used in amount0, but
its name in main was 'spec', doubly confusing. aname is the attachname
to present to the server, so use that everywhere.

### DIFF
--- a/sys/src/cmd/mount.c
+++ b/sys/src/cmd/mount.c
@@ -16,13 +16,13 @@ void	catch(void *c, char*);
 
 char *keyspec = "";
 
-int
-amount0(int fd, char *mntpt, int flags, char *aname, char *keyspec)
+static int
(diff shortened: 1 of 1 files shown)


rminnich:
tested on boot and by hand and it seems to work

floren (approved):

PR #995, opened 2020-09-10
Support basic open/read/close/clunk for 9P2000.L files

rminnich:
With this change and a 9P2000.L server running at 192.168.0.1:8080

192.168.0.8# srv tcp!192.168.0.1!8080 x
post...

192.168.0.8# mount /srv/x /n/x
mount: mount /n/x: nonsense value of msize in fversion

192.168.0.8# mount -M N /srv/x /n/x

192.168.0.8# cat /n/x/shosts
errno 2
mntrpcread: convM2S failed
cat: can't open /n/x/shosts: '/n/x/shosts' mount rpc error

192.168.0.8# cat /n/x/hosts
127.0.0.1       localhost
127.0.1.1       rminnich-MacBookPro

::1     ip6-localhost ip6-loopback
fe00::0 ip6-localnet
ff00::0 ip6-mcastprefix
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters

192.168.0.1 centre
192.168.0.2 up
192.168.0.3 api2 u00:07:32:4b:f9:f3
192.168.0.4 api3 u00:07:32:4c:04:e6
192.168.0.5 api4 u00:07:32:4c:25:36
192.168.0.6 rome uc8:b3:73:1f:44:55
192.168.0.7 a apu2 u00:0d:b9:49:0f:04
192.168.0.8 z zero u00:00:00:00:00:00
192.168.0.239 router

Some progress anyway. This means at least we could read files
from ipfs.

No ls yet. Can't read directories.


sys/src/libc/port/errno2str.c
```diff
@@ -0,0 +1,155 @@
...
+ * modified, propagated, or distributed except according to the terms contained
+ * in the LICENSE file.
+ */
+
+#include <u.h>
+#include <libc.h>
+
+static char *errors[] = {
+	[1] = "operation not permitted",
+	[2] = "no such file or directory",
```
fhs, line 15: What system are these error strings from? Maybe we should match them to what Go expects on Plan 9:
```
$ go doc -all syscall | grep ENOENT
	ENOENT          = Errno(0x2)
$ GOOS=plan9 go doc -all syscall | grep ENOENT
	ENOENT       = NewError("file does not exist")
```

rminnich: They are from the p9 server. 
I believe they come from Linux directly, which I figure is OK. This is definitely the unix text. 

What really ought to happen is we use the constants in libc. I'll take a look.

rminnich: OK, see what you think, I lined them up I think.

sys/include/libc.h
```diff
@@ -599,6 +599,7 @@ extern	int	chdir(const char*);
 extern	int	close(int);
 extern	int	create(const char*, int, uint32_t);
 extern	int	dup(int, int);
+extern  char*   errno2str(uint errno);
```
floren, line 602: I know we don't tend to make a big fuss about formatting but...

rminnich: tell GitHub to stop displaying this incorrectly. It lines up right if you look at the raw file.

rminnich: anyway, fixed, I screwed up and did not use tabs. My bad.

floren (dismissed):

floren (approved):

gmacd (approved):

PR #1000, opened 2020-09-12
Fix crash in mount syscall when dc is 0

### LINKED ISSUE #998: syscall.Mount crashes in Go
#### Steps to reproduce
Build this program in unix and run in harvey:
```Go
package main
import "syscall"
func main() {
	if err := syscall.Mount(0, 0, "", 0, ""); err != nil {
		println(err.Error())
	}
}
```
To build: `GOOS=plan9 go build mount.go`

#### Expected behavior

Program should print an error. On 9front:
```
cpu% ./mount
inappropriate use of fd
```

#### Actual behavior

Program crashes:
```
cpu% ./mount
mount 1063: suicide: invalid address 0x0/1 in sys call pc=0x25d980
```

fhs:
When dc is 0, we try to do a bind, but since the `name` argument is nil,
we crash with `invalid address`. The fix falls back to 'M' in this case,
which also helps plan9 programs that doesn't know about dc.

Fixes #998

### DIFF
--- a/sys/src/9/port/sysfile.c
+++ b/sys/src/9/port/sysfile.c
@@ -1248,9 +1248,9 @@ sysmount(Ar0* ar0, ...)
 	va_start(list, ar0);
 
 	/*
-	 * int mount(int fd, int afd, char* old, int flag, char* aname);
+	 * int mount(int fd, int afd, char* old, int flag, char* aname, int dc);
 	 * should be
-	 * long mount(int fd, int afd, char* old, int flag, char* aname);
+	 * long mount(int fd, int afd, char* old, int flag, char* aname, int dc);
 	 */
 	fd = va_arg(list, int);
 	afd = va_arg(list, int);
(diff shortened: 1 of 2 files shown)


rminnich (dismissed):
Nice catch. This was not an issue back when we did not support existing binaries. This is a really reasonable alternative to adding another system call. My only worry centers on when dc is garbage with old programs that don't have this new API?

Do you think we should have a deprecated _mount and then the new mount, as was done with _brk?

I'm going to approve this, up to you what we should do. note that in old plan 9 ABI all args are on the stack, not in registers as in the new ABI.

fhs:
> My only worry centers on when dc is garbage with old programs that don't have this new API?

That is a concern. I think a new syscall should be added.

fhs:
Added `mount_` syscall.

sys/src/9/port/sysfile.c
```diff
@@ -1260,9 +1260,35 @@ sysmount(Ar0* ar0, ...)
 	dc = va_arg(list, int);
 	va_end(list);
 
+	if(dc == 0){
```
rminnich, line 1263: I think we should not do this one part here? It should just be an error I think?

rminnich (commented):
This is really good. It's exactly the right thing. My only quibble: I think if 'dc' is invalid that's an error. There should be no default. Or do you want the behavior to be 'use 0 to get a default'?

sys/src/9/port/sysfile.c
fhs: Changed it to an error.

rminnich (approved):

PR #1001, opened 2020-09-14
readdir for 9p2000.L

rminnich:
Working read of 9P2000.L directories. Note that you can cat
but not ls -l yet, as we don't have stat, and the first thing
ls does is stat the directory.

This works with a 9P2000.L server running at 192.168.0.1!8080
srv tcp!192.168.0.1!8080 x
mount -M N /srv/x /n/x
cat /n/x

### DIFF
--- a/sys/src/libc/9sys/convM2S.c
+++ b/sys/src/libc/9sys/convM2S.c
@@ -298,6 +298,19 @@ convM2S(uint8_t *ap, uint nap, Fcall *f)
 		p += f->count;
 		break;
 
+	case Rreaddir:
+		if(p+BIT32SZ > ep) {
+			return 0;
+		}
+		f->count = GBIT32(p);
+		p += BIT32SZ;
+		if(p+f->count > ep) {
+			return 0;
+		}
+		f->data = (char*)p;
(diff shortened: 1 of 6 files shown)


sys/src/libc/9sys/convM2S.c
```diff
@@ -298,6 +298,21 @@ convM2S(uint8_t *ap, uint nap, Fcall *f)
 		p += f->count;
 		break;
 
+	case Rreaddir:
+		if(p+BIT32SZ > ep) {
+			print("rreaddir: p %p ep %p damn\n", p, ep);
```
gmacd, line 303: should be readdir?

gmacd (approved):

fhs (dismissed):

sys/src/libc/9sys/convM2S.c
gmacd, line 303:
```suggestion
			print("readdir: p %p ep %p damn\n", p, ep);
```

rminnich:
PTAL

gmacd (approved):

PR #1002, opened 2020-09-20
stat for 9P2000.L / devmntn

rminnich:
Sample session with a 9P2000.L server (e.g. ipfs):
srv tcp!192.168.0.1!8080 x
mount -M N /srv/x /n/x
ls -l /n/x

9P2000.L has two big problems: first, a directory read does not return
stat records, as it should; second, the stat has fields missing, particularly
the file name, a string for uid, and a string for gid.

These are hard to work around. The result in output like this:

% ls -l/n/x

------w-r--  0 '' '' 0 Feb  7  2006 /n/x/util

because a readdir does not return stat records, we are missing user, group, size information.

But:
192.168.0.7# ls -ld /n/x
d-rwxrwxr-x N 31 1000 1000 4096 Feb  7  2006 /n/x

Because in this case, the code does a stat on /n/x and gets a stat record back.

Now for the first case, what has to happen, but not in this PR, is on a readdir,
the code must do a full walk and stat *for every directory entry*. This means the
number of RPCs for a directory is 2x the number of directory entries, instead
of 1 total as it is in 9P2000. 64 directory entries will result in 129 RPCs
in 9P2000.L and 1 in 9P2000 (129: 1 for readdir, 128 for 2 RPCs for each entry).

Further, on a stat, the file name is not in the stat; we have to pull the name
from the channel. What happens if the name changed after the initial walk?
You'll have the wrong name in a stat. That's not great.

9P2000.L violates plan 9 rules in other small ways, such as including an inumber
in the stat, because few if any real Plan 9 systems have inumbers, but
Linux had its own requirements.

It does get time right. Plan 9 has a 32-bit time problem; 9P2000.L does not.

It's worth supporting 9P2000.L with all its defects, so here we take
another step on that journey.

The code in devmntn.c:mntstat is nasty. I'd just as soon it stayed that
way, rather than bring too much of 9P2000.L's non-plan9-like nature
into fcall.h and libc.

### DIFF
--- a/sys/include/fcall.h
+++ b/sys/include/fcall.h
@@ -97,6 +97,30 @@ struct	Fcall
 #define	NOFID		(uint32_t)~0U	/* Dummy fid */
 #define	IOHDRSZ		24	/* ample room for Twrite/Rread header (iounit) */
 
+// 9P2000.L, for some reason, lets callers closely tune what comes from a gettattr.
+// This *may* be because HPC file systems make getting at some types of metadata
+// expensive. Or it may be a pointless over-optimization; we suspect the latter.
+// Just ask for it all. Bandwidth is free.
+
+#define P9_GETATTR_MODE         0x00000001ULL
+#define P9_GETATTR_NLINK        0x00000002ULL
+#define P9_GETATTR_UID          0x00000004ULL
+#define P9_GETATTR_GID          0x00000008ULL
+#define P9_GETATTR_RDEV         0x00000010ULL
+#define P9_GETATTR_ATIME        0x00000020ULL
+#define P9_GETATTR_MTIME        0x00000040ULL
+#define P9_GETATTR_CTIME        0x00000080ULL
+#define P9_GETATTR_INO          0x00000100ULL
+#define P9_GETATTR_SIZE         0x00000200ULL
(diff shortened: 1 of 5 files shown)


sys/log/telnet
```diff
@@ -0,0 +1,2 @@
+192.168.0.7 Sep 14 02:43:01 (192.168.0.1!56240) failed as SSH-2.0-Go: auth server protocol botch
```
gmacd, line 1: should this file be removed from git and added to gitignore?

gmacd (dismissed):
There's definitely a bit of trust  here - if it works, it's good.

sys/include/fcall.h
```diff
@@ -97,6 +97,30 @@ struct	Fcall
 #define	NOFID		(uint32_t)~0U	/* Dummy fid */
 #define	IOHDRSZ		24	/* ample room for Twrite/Rread header (iounit) */
 
+// 9P2000.L, tor some reason, lets callers closely tune what comes from a gettattr.
```
fhs, line 100:
```suggestion
// 9P2000.L, for some reason, lets callers closely tune what comes from a gettattr.
```

rminnich:
now if only we could solve the bootstrap.sh mess

gmacd (approved):

PR #1008, opened 2020-09-24
Fix embarrasing errors in ramfs

rminnich:
the read was not doing the right # of bytes.

Debug prints are now disabled.


rminnich:
192.168.0.7# cp /lib/ndb/local '#@'
192.168.0.7# cat '#@/local'

#
#  files comprising the database, use as many as you like, see ndb(6)
#
database=
        file=/lib/ndb/local
        file=/lib/ndb/common

auth=sources.cs.bell-labs.com authdom=outside.plan9.bell-labs.com

#
#  because the public demands the name localsource
#
ip=127.0.0.1 sys=localhost dom=localhost
192.168.0.7# 

now works

gmacd (approved):

PR #1019, opened 2020-10-13
port/page.c cleanup

rminnich:
In the early days we had serious toolchain issues but they
seem to be fixed. This code was probably written to get around
the early STD or clang implicit float issues.

Tested with clang-10 and gcc.


rminnich:
you probably want to test though it does work for me.

elbing (approved):

elbing:
Someone should review the install deps of Ubuntu 18.04 on CI.

PR #1020, opened 2020-10-13
a lousy attempt to get the I217 to work - draft and junk, don't merge

orangecms:


sys/src/9/amd64/ether82563.c
```diff
@@ -1923,7 +1926,13 @@ i82563reset(Ctlr *ctlr)
 	}
 	if (r != 0 && r != 0xBABA){
 		print("%s: bad EEPROM checksum - %#.4ux", tname[type], r);
-		if (type != i82579 && type != i210 && type != i217 && type != i218 && type != i219) {
+		if (
+      type != i82579 &&
```
rminnich, line 1930: I will say I'm not a fan of this wrapping. It defeats grep.

orangecms: Interesting point. Quite often, I'm using `grep -C 3` etc to get more context. Can you elaborate a bit what issue you see?

PR #1021, opened 2020-10-13
CI: retrieve keys for LLVM repo

orangecms:

### DIFF
--- a/.github/workflows/harvey.yml
+++ b/.github/workflows/harvey.yml
@@ -24,10 +24,11 @@ jobs:
     - uses: actions/checkout@v2
     - name: install deps
       run: |
-        sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
-        sudo add-apt-repository 'deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-10 main' -y
+        sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y # see https://wiki.ubuntu.com/ToolChain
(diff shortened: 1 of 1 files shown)


.github/workflows/harvey.yml
```diff
@@ -25,9 +25,8 @@ jobs:
     - name: install deps
       run: |
         sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
-        sudo add-apt-repository 'deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-10 main' -y
         sudo apt-get update
-        sudo apt-get install -y --force-yes ${{ matrix.cc }} rc qemu go-dep
+        sudo apt-get install -y ${{ matrix.cc }} rc qemu go-dep
```
gmacd, line 29: do you need to install  ${{ matrix.cc }}?

gmacd: never mind, will be useful if we need to install newer/other versions

gmacd (approved):

.github/workflows/harvey.yml
gmacd: also won't do anything if the compiler is already installed, so it's fine.

fhs (approved):
Do we need to add back the `--force-yes` flag now that we're staying on Ubuntu 18.04? The commit message needs to be updated otherwise.

orangecms:
yea I'll make it "will be" ;)

PR #1022, opened 2020-10-13
ci fix plus plus

orangecms:
- CI: upgrade to Ubuntu 20.04 Focal
- port/page.c cleanup


orangecms:
Result of the experiments: We can't just upgrade to 20.04. The issue *might* be with QEMU; needs investigation. Closing for now.

PR #1023, opened 2020-10-13
ci fix junkectomy

orangecms:
This includes both of these patches:

- CI: retrieve keys for LLVM repo
- port/page.c cleanup

Please CI be nice and let them pass. :)


orangecms:
no longer needed - back to the original PR #1019

PR #1025, opened 2020-10-20
uartpci: add support for WCH CH382 uart pci card

gmacd:
Uart port offset found in linux.
Initialise devuart first to allow early connection to pci uart.
Can initialise by adding the following after the uart has been created in uartpci.c:
  i8250uartconsole(uart, nil);
(Would be better to drive it off the kernel args or some other cfg, but no good ideas yet)

### DIFF
--- a/sys/src/9/386/uartpci.c
+++ b/sys/src/9/386/uartpci.c
@@ -20,14 +20,14 @@ extern PhysUart pciphysuart;
 extern void* i8250alloc(int, int, int);
 
 static Uart*
-uartpci(int ctlrno, Pcidev* p, int barno, int n, int freq, char* name)
+uartpci(int ctlrno, Pcidev* p, int barno, int first_offset, int n, int freq, char* name)
 {
 	int i, io;
 	void *ctlr;
 	char buf[64];
 	Uart *head, *uart;
 
-	io = p->mem[barno].bar & ~0x01;
+	io = (p->mem[barno].bar & ~0x01) + first_offset;
 	snprint(buf, sizeof(buf), "%s%d", pciphysuart.name, ctlrno);
 	if(ioalloc(io, p->mem[barno].size, 0, buf) < 0){
 		print("uartpci: I/O 0x%X in use\n", io);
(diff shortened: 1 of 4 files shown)


orangecms (approved):
All fine to me.

> WCH CH353 (vid not in pci db) 

Yea well... 🤐 searching for it gives `4348`/`3453`. Vendors and stuff, I suppose...

gmacd:
My mistake, it's a CH382:
https://github.com/torvalds/linux/blob/master/drivers/tty/serial/8250/8250_pci.c#L1904

sys/src/9/386/uartpci.c
```diff
@@ -20,14 +20,14 @@ extern PhysUart pciphysuart;
 extern void* i8250alloc(int, int, int);
 
 static Uart*
-uartpci(int ctlrno, Pcidev* p, int barno, int n, int freq, char* name)
+uartpci(int ctlrno, Pcidev* p, int barno, int first_offset, int n, int freq, char* name)
```
rminnich, line 23: first_offset? why is it called that? What does that name mean?

n is the number of uarts. Will first_offset be different for every value of n? As this code is written, it is always the same. Is this code written in a way that it only works if n is 1? If so, that ought to be documented.

You're changing the uartpci API for all callers to accommodate a quirk of one part. Is there no other way to do this? I can't think of one, just asking. Why is the value c0? 

You mention the vid is not in the pic db -- why don't we just add it and then remove this comment?

gmacd: first_offset could be just offset.  the name was taken from the equivalent linux code.

assuming each uart has an initial offset (bar + first_offset), each subsequent uart in the iteration is 8 bytes after.  it's the initial offset for the sequence.  so n=0 has offset+0, n=1 has offset+8, n=2 has offset+16, etc.

no idea why it's c0 - that's what linux uses and it works.  not sure of a better way to do it.

when i say the vid isn't in the pci db, I meant the PCI ID repo (https://pci-ids.ucw.cz).  there's no vid for 0x1c00.  WCH have a vid of 0x4348 which they use for some cards, but they seem to use 0x1c00 here.

outside of this, we should probably update our coy of the PCI repo, assuming it's the same one.

rminnich (approved):

PR #1026, opened 2020-10-21
amd64/main.c: add a post function to improve bringup debugging

rminnich:
harvey does not boot with ease on many modern x86 CPUs yet.
Add a post() function for main to provide copious debug
output, if a static variable is set.

### DIFF
--- a/sys/src/9/amd64/main.c
+++ b/sys/src/9/amd64/main.c
@@ -62,7 +62,31 @@ static int vflag;
 
 int nosmp;
 int acpionly = 1;
+static int showpost = 0;
 
+// Harvey is out of date on many platforms, and frequently the only
+// indication is a brick.
+// Modern systems have extensive logging of outb to port 0x80, as well
+// as serial over lan support. Further, the emulated serial easily runs
+// as fast as the x86 can push it over LPC.
+// This post function is intended only for use in main(), when we are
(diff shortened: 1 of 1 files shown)


sys/src/9/amd64/main.c
```diff
@@ -541,9 +566,9 @@ main(uint32_t mbmagic, uint32_t mbaddress)
 	active.exiting = 0;
 	active.nbooting = 0;
 
-	asminit();
-	multiboot(mbmagic, mbaddress, 0);
-	options(oargc, oargv);
+	asminit(); post("	asminit();", postterminal++);
+	multiboot(mbmagic, mbaddress, 0); post("	multiboot(mbmagic, mbaddress, 0);", postterminal++);
```
gmacd, line 570: What's up with the tabbing here? Intentional?

gmacd (approved):

sys/src/9/amd64/main.c
rminnich: yeah, it's so via grep I can look for only those lines.

rminnich:
I've used this just now on an AMD Zen CRB and ... it is a timesaver to say the least.

orangecms:
Nice! I need to try it on the A300 :)

PR #1031, opened 2020-10-24
remove the old kernel exec code

rminnich:
This was never done to success, the goal was to exec
a process in kernel mode.


rminnich:
This was part of a ph thesis that remained unfinished. I don't remember it ever being compiled in past about 2015.

elbing (approved):

PR #1032, opened 2020-10-25
ahci: update ahci identification and init, remove support for esb

gmacd:
sdiahci was hanging when trying to intialise on a Skylake i7-6700K,
when calling ahciidle for port 5.  This appears to have been caused
by writing to 0x90 in the PCI register.  Commenting it out removes
the hang (though I haven't tested beyond that).

There seems to be a bit of init code being called for almost all
Intel controllers - this probably wasn't the original intention.
Given that the PCI registers would change from controller to
controller, this is probably not the best way to do it.  Given
that the code is quite old and therefore the controllers also,
it might be best to remove controller specific init code, follow
the AHCI spec, and see how far we can get.

I restructured the code to put more of the initialisation code in
one place.  This makes it a bit easier to see the same steps being
carried out multiple times.

### DIFF
--- a/sys/src/9/amd64/sdiahci.c
+++ b/sys/src/9/amd64/sdiahci.c
@@ -58,22 +58,13 @@ enum {
 	Maxintrspertick = 2000,		/* was 1000 */
 };
 
-/* pci space configuration */
 enum {
-	Pmap	= 0x90,
-	Ppcs	= 0x91,
-	Prev	= 0xa8,
-};
-
-enum {
-	Tesb,
 	Tich,
 	Tsb600,
 	Tunk,
 };
 
 static char *tname[] = {
-	"63xxesb",
 	"ich",
 	"sb600",
 	"unknown",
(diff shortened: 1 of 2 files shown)


rminnich (commented):
what do 9legacy and 9front do?

gmacd:
Working out what 9front and 9legacy support might take a bit more time than I have right now - I assumed they didn't support anything newer than 10 years old, but that might not be correct.

Looking at the linux code though, there's only one obviously related place where the PCI registers are written to: https://github.com/torvalds/linux/blob/3650b228f83adda7e5ee532e2b90429c03f7b9ec/drivers/ata/ahci.c#L1621

This corresponds to our intel-specific code to enable the ports:
```
pcicfgw16(ctlr->pci, 0x92, pcicfgr16(ctlr->pci, 0x92) | 0xf);
```

So according to linux, it looks like that should be applied for intel device IDs 0x19xx (not all 0x19xx though) (which is actually none of those we directly support).

rminnich (changes requested):
if you want to commit be sure to take the WIP: reference out, and it would be nice to have a better explanation of what this does, it's more than just tidying.

rminnich:
never mind the comment, your explanation is fine.

rminnich (approved):
it's fine, don't know what I was thinking.

gmacd:
Thanks, though I do plan on tidying it - it’s still a bit scrappy.

sys/src/9/amd64/sdiahci.c
```diff
@@ -818,38 +809,30 @@ countbits(uint32_t u)
...
 
-	if((u&Hsam) == 0)
-		h->ghc |= Hae;
+	if (Intel(ctlr)) {
+		// Port control and status register
+		// Present on ICH9
+		// Not present on 100 series
+		// Enable all ports (also OOB failure retries)
+		/* configure drives 0-5 as ahci sata  (c.f. errata) */
+		pcicfgw16(ctlr->pci, 0x92, pcicfgr16(ctlr->pci, 0x92) | 0xf);
```
rminnich, line 823: I worry a bit here, given Intel's penchant for moving registers around ... should you check the devid?

rminnich (approved):

sys/src/9/amd64/sdiahci.c
gmacd: yeah...  on my controller, the register had been removed.  I'd been considering lifting the devid and some of the categorisation from the linux ahci.c code.  That can be used for this purpose.

rminnich:
graham, if this works for you, push it, we'll work out issues later.

rminnich (approved):

PR #1046, opened 2020-12-22
Clean up setlabel/gotolabel

dancrossnyc:
Get rid of `STACKPAD` in syscall.c.


rminnich (approved):
This is one of the Original Sins of NIX/harvey. It's so nice to see it cleaned up.

Tested on APU2, works fine; t510, crashes in the same way. go for it.

PR #1047, opened 2020-12-22
Remove a host of 32-bit assumptions

dancrossnyc:
There are many, many more lurking in the code.


rminnich (approved):
tested on t510 with 4g and 8g. and apu2. 

sadly, I still get the fault on the 8g system, but this makes things no worse.

PR #1048, opened 2020-12-23
clang-format: start to clean up code

dancrossnyc:
Code has gotten inconsistent.  Start to clean it up.


rminnich (approved):
looks good. We may want to run this across the tree?

dancrossnyc:
> looks good. We may want to run this across the tree?

Indeed, but it'll be a big diff and I wanted to just send this initially to make sure it was ok with folks.

PR #1049, opened 2020-12-23
The Great Reformatting

dancrossnyc:
Run `clang-format` across the kernel.


rminnich (approved):
works for me. I did not even look at them all; I can live with whatever clangfmt does. The beauty of it is we can all do this again later if clangfmt gets better.

rminnich:
tested on apu2, works fine with gcc and clang

PR #1054, opened 2020-12-28
Alternative kernel build.json (allbuild.json) that adds uroot archive to ramdisk and extracts in cpurc

gmacd:
If you build `sys/src/9/amd64/allbuild.json`, then run cpu, cpurc will extract the uroot archive to the ramdisk and bind it to root.


rminnich (commented):
I'd say copy build.json to allbin.json and this doesn't even need to be a WIP.

rminnich (approved):
good enough for now. You can still use u-root to combine those two commands to one but that can wait.

rminnich (approved):

PR #1056, opened 2021-01-01
Clean up mmuwalk.

dancrossnyc:
It didn't terminate correctly.

### DIFF
--- a/sys/src/9/amd64/mmu.c
+++ b/sys/src/9/amd64/mmu.c
@@ -161,7 +161,7 @@ allocapage(void)
 		char *pp = mallocalign(npage * PTSZ, PTSZ, 0, 0);
 		assert(pp != nil);
 		p = pp + (npage / 2) * PTSZ;
-	} else {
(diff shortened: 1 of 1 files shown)


rminnich (approved):
apu2 still works fine, but t510 eatmem test still panics.

Actually, the apu2 eatmem test still panics with a different problem, likely fixable but still ..
192.168.0.7# regress/kernel/eatmem

panic: cpu0: kickpager(): no physical memory
cpu0 stack: k:ffff8000002a730f k:ffff8000002dd0f5 k:ffff8000002dbe45 k:ffff8000002d5f9d k:ffff8000002d5abc k:ffff800000252539 k:ffff80000025301c k:ffff800000252e0a k:ffff80000022c863
die:wait forever

so one issue with removing kseg0 is that user programs can now run the kernel out of memory I guess.

dancrossnyc:
Removing kseg0 doesn't cause that problem, though it may exacerbate it.

PR #1057, opened 2021-01-01
Add a walk entry to devregress to test VM mappings

rminnich:
The walk entry lets us see how things are mapped, e.g.
192.168.0.7# echo -n 0xffff800000000000>'#R/walk'
cpu0: pml4 0xffff8000034c5000
cpu0: mmu l3 pte 0xffff8000034c5800 = 111023
cpu0: mmu l2 pte 0xffff800000111000 = 1602023
cpu0: mmu l1 pte 0xffff800001602000 = 1606023
cpu0: mmu l0 pte 0xffff800001606000 = 8000000000000003
192.168.0.7#

This whould help us debug some current problems, and also
understand how things end up being mapped.

### DIFF
--- a/sys/src/9/port/devregress.c
+++ b/sys/src/9/port/devregress.c
@@ -13,6 +13,7 @@ enum {
 	Qmalloc,
 	Qtsleep,
 	Qlock,
+	Qwalk,
 	Qmax,
 };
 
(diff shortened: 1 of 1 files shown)


rminnich:
Also fixed a bug in devregress

gmacd (approved):

sys/src/9/port/devregress.c
```diff
@@ -126,6 +129,12 @@ regresswrite(Chan *c, void *a, i32 n, i64 offset)
 			error("Only v or V");
 		return n;
 
+	case Qwalk:
+		p = a;
+		addr = strtoull(p, 0, 0);
```
dancrossnyc, line 134: It may be useful to validate that this is actually a kernel address here. Something like,

    if (addr < KZERO) error("Invalid kernel address");

rminnich: I thought about that but as part of regression, we might want to, rather, let it try to walk it and hope it fails?

dancrossnyc: Ok; sounds fine to me.

PR #1059, opened 2021-01-01
Stop using go modules. They don't work yet.

rminnich:


rminnich:
People can do their own go get if they have to. This situation is not working for modules just now.

fhs:
Module usage already removed by https://github.com/Harvey-OS/harvey/pull/1065

PR #1060, opened 2021-01-02
Fix mapping error

dancrossnyc:
As we map each page of the address space, the amount
of address space we need to map decreases by the page
size we just mapped.


rminnich (approved):
t510 now boots. apu2 still boots.

eatmem still causes a panic, eventually, but still, things are far better.

PR #1061, opened 2021-01-02
WIP: Have kernel init support more service types

gmacd:

### DIFF
--- a/sys/src/9/boot/boot.c
+++ b/sys/src/9/boot/boot.c
@@ -70,7 +70,7 @@ acpiirq(void)
 void
 boot(int argc, char *argv[])
 {
-	int fd, afd, srvt;
+	int fd, afd;
 	Method *mp;
 	char *cmd, cmdbuf[64], *iargv[16];
 	char rootbuf[64];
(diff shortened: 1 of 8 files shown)


rminnich (commented):
looks good, I wonder why it broke ...

sys/src/9/boot/boot.c
```diff
@@ -224,17 +224,9 @@ boot(int argc, char *argv[])
...
-		if(!srvt){
-			sprint(cmdbuf, "/%s/bin/init -%s%s", cputype,
-			       "t", mflag ? "m" : "");
-			cmd = cmdbuf;
-		} else {
-			sprint(cmdbuf, "/%s/bin/init -%s%s", cputype,
-			       "c", mflag ? "m" : "");
-			cmd = cmdbuf;
-		}
+	if(cmd == nil && service != nil) {
```
rminnich, line 227: The easiest fix for this:
if(cmd == nil && service[0] != 0 {

rminnich: The bigger problem is this should not have built for you, which many mean we're missing a flag.

rminnich: ah, it instantly fails for me with clang:
boot.c:227:19: error: comparison of array 'service' not equal to a null pointer is always true [-Werror,-Wtautological-pointer-compare]
        if(cmd == nil && service != nil) {
                         ^~~~~~~    ~~~
1 error generated.
exit status 1

but stupid gcc does not catch it. Crap.

rminnich: This one change will fix it and do what you want
index 8a763de25..e483ce438 100644
--- a/sys/src/9/boot/boot.c
+++ b/sys/src/9/boot/boot.c
@@ -224,7 +224,7 @@ boot(int argc, char *argv[])
                close(afd);
 
        cmd = getenv("init");
-       if(cmd == nil && service != nil) {
+       if(cmd == nil && service[0] != 0) {
                sprint(cmdbuf, "/%s/bin/init -s %s %s", cputype, service, mflag ? "m" : "");
                cmd = cmdbuf;
        }

it's quite nice to have all that stuff there.

PR #1062, opened 2021-01-03
Add back [DEFG]S registers to Ureg

### LINKED ISSUE #999: nil pointer in Go doesn't generate a panic
#### Steps to reproduce
Run this program in harvey:
```Go
package main

func main() {
	var p *int
	println(*p)
}
```
Built on linux using Go 1.15.2: `GOOS=plan9 go build nil.go`

#### Expected behavior

It should generate a go panic. On 9front:
```
cpu% ./nil
panic: runtime error: invalid memory address or nil pointer dereference
[signal sys: trap: fault read code=0x0 addr=0x0 pc=0x258d44]

goroutine 1 [running]:
main.main()
	/home/fhs/tmp/nil.go:5 +0x24
```

#### Actual behavior

```
cpu% ./nil
nil 173: suicide: sys: trap: fault read addr=0x23 pc=0x2295cf
```

fhs:
Commit 6ab2e16c9f1058aab427f79d4516d281c8e6039b removed these
segmentation registers, but we still need them for Go programs to work
correctly.

The Go runtime uses ureg.sp when it needs to print a stack trace after
the program dereferences a nil pointer and receives a note from the
kernel. Since sp comes after the segmentation registers in the struct,
the sp offset within the struct would be wrong if the segmentation
registers are missing.

After this change, Go programs will now panic (and print a stack trace)
instead of crashing when dereferencing a nil pointer.

Fixes #999

### DIFF
--- a/amd64/include/ureg.h
+++ b/amd64/include/ureg.h
@@ -24,13 +24,13 @@ struct Ureg {
 	uint64_t	r14;
 	uint64_t	r15;
 
-  // these are pointless and I don't want to waste a second on them.
-  // I'm sure it will break something but let's see.
-  // Not pushing them simplifies the interrupt handler a lot.
-  //	uint16_t	ds;
-  //	uint16_t	es;
-  //	uint16_t	fs;
-  //	uint16_t	gs;
+	// These are pointless and not pushing them would simplify the
+	// interrupt handler a lot. However, removing them would break
+	// compatibility with Go Plan 9 binaries.
+	uint16_t	ds;
+	uint16_t	es;
+	uint16_t	fs;
+	uint16_t	gs;
 
 	uint64_t	type;
 	uint64_t	error;				/* error code (or zero) */
--- a/sys/src/9/amd64/l64idt.S
+++ b/sys/src/9/amd64/l64idt.S
@@ -85,6 +85,7 @@ _intrcommon:
 	SWAPGS
 
 _intrnested:
+	SUBQ	$8, %rsp				/* [DEFG]S */
 	PUSHQ	%r15
 	PUSHQ	%r14
 	PUSHQ	%r13
@@ -126,6 +127,7 @@ _intrr:
 	POPQ	%r13
 	POPQ	%r14
 	POPQ	%r15
+	ADDQ	$8, %rsp				/* [DEFG]S */
 
 	CMPW	$SSEL(SiCS, SsTIGDT|SsRPL0), 24(%rsp) 	/* old CS */
 	JE	_iretnested
--- a/sys/src/9/amd64/l64vsyscall.S
+++ b/sys/src/9/amd64/l64vsyscall.S
@@ -50,7 +50,7 @@ syscallentry:
 
 	movq	%gs:32, %r15 	/* restore r15 from m->rathole */
 
-	SUBQ	$(17*8), %rsp
+	SUBQ	$(18*8), %rsp
 
 	MOVQ	%rAX, (0*8)(%rsp)	// ureg.ax
 	MOVQ	%rBX, (1*8)(%rsp)	// ureg.bx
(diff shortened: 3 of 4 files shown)


dancrossnyc (commented):
Do you actually need to save/restore them? Right now, it appears that you're just putting some space for them on the stack, but not otherwise touching them.

fhs:
> Do you actually need to save/restore them? Right now, it appears that you're just putting some space for them on the stack, but not otherwise touching them.

I'm going by what we had before we [removed them](https://github.com/Harvey-OS/harvey/commit/6ab2e16c9f1058aab427f79d4516d281c8e6039b). We didn't bother saving them before either.

dancrossnyc:
I guess I don't quite understand why it matters; all of this happens once we're _in the kernel_; I guess the older plan9 style syscall handlers require a different layout for ureg? Why? Are those 32-bit binaries (sorry; I'm not familiar with what's been copied in).

It seems like it would be more fruitful to find whatever assumptions the a.out code is making about the existence (or lack thereof) of x86 segmentation registers and fix that.

fhs:
> It seems like it would be more fruitful to find whatever assumptions the a.out code is making about the existence (or lack thereof) of x86 segmentation registers and fix that.

We can't fix the a.out code even if we want to because these are Go programs and we need to match other Plan 9 systems that Go targets (9legacy and 9front).

I don't think Go programs actually uses the segmentation registers, but it seems to be using `ureg.sp` when it needs to print a stack trace after the program dereferences a nil pointer and receives a note from the kernel. Unfortunately, `sp` comes *after* the segmentation registers in the struct, so the `sp` offset within the struct would be wrong if the segmentation registers are missing.

rminnich (approved):

dancrossnyc:
> > It seems like it would be more fruitful to find whatever assumptions the a.out code is making about the existence (or lack thereof) of x86 segmentation registers and fix that.
> 
> We can't fix the a.out code even if we want to because these are Go programs and we need to match other Plan 9 systems that Go targets (9legacy and 9front).

That doesn't make a lot of sense to me. Consider that the Ureg is something that's created by the kernel on entry to the kernel and only used in the kernel; userspace shouldn't have any visibility into that, and shouldn't care what the kernel does as long as it faithfully restores the userspace environment on return. In other words, the fact that these are Go binaries running in userspace shouldn't a priori dictate how the kernel lays out its own data structures.

> I don't think Go programs actually uses the segmentation registers, but it seems to be using `ureg.sp` when it needs to print a stack trace after the program dereferences a nil pointer and receives a note from the kernel. Unfortunately, `sp` comes _after_ the segmentation registers in the struct, so the `sp` offset within the struct would be wrong if the segmentation registers are missing.

But that's back in userspace, right? Note that Ureg here is being constructed on the proc's _kernel_ stack.

What this suggests to me is that the a.out code has a different idea of they layout of Ureg than the rest of the kernel, which is honestly not great. Instead, if we need to do some kind of translation to whatever the note handler is expecting the kernel to give it, we should do that at the point where that structure is created.

fhs:
> What this suggests to me is that the a.out code has a different idea of they layout of Ureg than the rest of the kernel, which is honestly not great. Instead, if we need to do some kind of translation to whatever the note handler is expecting the kernel to give it, we should do that at the point where that structure is created.

I think I understand your point. It doesn't seem like a small change since the kernel uses Ureg defined in `/amd64/include/ureg.h` everywhere. We'd need to define a separate Ureg within the kernel (that doesn't have the segmentation registers) but send the one defined in `/amd64/include/ureg.h` (which needs to have the segmentation registers) to note handlers.

dancrossnyc:
> > What this suggests to me is that the a.out code has a different idea of they layout of Ureg than the rest of the kernel, which is honestly not great. Instead, if we need to do some kind of translation to whatever the note handler is expecting the kernel to give it, we should do that at the point where that structure is created.
> 
> I think I understand your point. It doesn't seem like a small change since the kernel uses Ureg defined in `/amd64/include/ureg.h` everywhere. We'd need to define a separate Ureg within the kernel (that doesn't have the segmentation registers) but send the one defined in `/amd64/include/ureg.h` (which needs to have the segmentation registers) to note handlers.

What I'm suggesting is that the kernel has to understand that a particular program is using a particular ABI (e.g., the ABI used by a.out plan9 binaries is different than SYSV/ELF). If the kernel is sending a note, which I suppose needs some information specific to registers vis a particular ABI, one should send the data required _by that ABI_. This would imply that sending a `Ureg` back to userspace for a.out binaries would be the responsibility of the part of the kernel that deals with a.out, instead of pushing that information everywhere in the kernel and having a Ureg that's the union of all Uregs required by all ABIs. For example, what if the upcall from the kernel used a different calling convention? This doesn't require a separate header file, but just a structure (or code) that puts the appropriate thing in the right place. The note delivery trampoline for a.out could do it.

PR #1067, opened 2021-01-09
Add devdc() and devname() functions to devmulti.c

michiboo:
Hi this is my first PR, just wanna keep it small at first.


orangecms:
Hi, thank you for contributing! For the changes to be accepted, you'll need to rebase this onto the current `main` upstream branch. The checks are currently failing because of the merge commits.
While you're at it: For an easier understanding, you  could directly reference the change in the summary line? Like, `devmulti: add devdc() and devname() functions`, and explain the idea behind it? Thank you!

rminnich (changes requested):
hi, this is not a kernel device, and the issue I raised is for kernel devices.

Please look at sys/src/9/*/*.c for some kernel devices.

rminnich:
I would actually suggest closing this PR, sorry!

PR #1068, opened 2021-01-12
ramfs: use Pages for backing store, not malloc

rminnich:
New hardware has a LOT of pages. If we unpack uroot into

Clearly we need to be a lot smarter than this code is, but it
already lets us get much further than we have so far.

### DIFF
--- a/sys/src/9/port/devramfs.c
+++ b/sys/src/9/port/devramfs.c
@@ -28,8 +28,11 @@ struct RamFile {
 	int perm;
 	int opencount;
 	int deleteonclose;
-	// A file is up to 4K 8192-byte blocks. That's 32M. Enough.
-	void *blocks[NBLOCK];
+	// A file is up to 16 pages. That's 32m.
+	// There is huge wastage here. But hey ...
+	// memory is so cheap. It's getting hard to buy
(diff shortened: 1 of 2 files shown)


rminnich:
This is obviously not the best ever, but it does let us experiment.

What we might do is create an allocator that has a pool of Pages and use that but, for now, this works.

sys/src/9/port/devramfs.c
```diff
@@ -301,12 +307,12 @@ ramreadblock(Chan *c, void *buf, i32 n, i64 off)
...
+	int fb = off >> 21, fl = (off + n) >> 21, offinblock = off & 0x1fffff;
 
 	// adjust read size and offset.
 	// we only read one block for now.
 	if(fl != fb){
-		n = 8192 - (off & 0x1fff);
+		n = (1<<21) - (off & 0x1fffff);
 	}
 
 	if(fb > NBLOCK)
```
gmacd, line 318: Should these comparisons `fb > NBLOCK` be replaced with `fb > 16`?
Or better, replace NBLOCK with NPAGES, set to 16.

rminnich:
we really ought to get cfs working for this.

PR #1071, opened 2021-01-24
Don't discard debug sections when linking the kernel

gmacd:


dancrossnyc (approved):
This is fine but just be aware that, depending on where the linker put them in the binary, they could potentially get blown away after loading. But if they're just there to make `objdump -S` easer to negotiate, that shouldn't matter.

gmacd:
Main reason was to help debugging in gdb, and viewing disassembly in objdump.  Any improvements that could be made?

Issue #1072, opened 2021-01-24
attempting to build harvey fails because there is no syslinux script in the utils folder


Lexi-kelly:
#### Expected behavior
it builds successfully
#### Actual behavior
"-bash: ./syslinux: No such file or directory"
#### Steps to reproduce
- open a terminal on ubuntu 18.04 and do sudo -i then run all commands as root
- follow getting started and building harvey exactly as it says
- skip running on qemu
- follow creating harvey image exactly
- watch it fail

#### QEMU output
<details>
```
<your output here>
```
</details>

fhs:
I'm guessing you're following the instructions here: https://github.com/Harvey-OS/harvey/wiki/Booting-Harvey-on-real-hardware-II-(USB) ?
These instructions don't get tested often as most of us netboot harvey these days.

The `syslinux` command is a Go program that's installed from [harvey-os.org/cmd/syslinux](https://pkg.go.dev/harvey-os.org/cmd/syslinux) into `$HARVEY/linux_amd64/bin`. It works if you run it from `util` directory, but we need to fix both the wiki instructions and how syslinux works.

Lexi-kelly:
yes I was following those instructions, if I installed the syslinux Go program would i then be able to build a bootable harvey USB?

fhs:
@Lexi-kelly I think that syslinux program should be deleted. Try the `util/build-img.sh` script instead from PR #1076.

Lexi-kelly:
doesn't the build-img script only build a qemu image? i'm looking to make a bootable usb image, is there a way i could change the script to do that or convert the qemu image to a bootable usb?

fhs:
@Lexi-kelly If you run `./util/build-img.sh -f raw` it'll produce a raw image instead of a qcow2. Then you can copy it to your usb stick.

Lexi-kelly:
thank you for your help with this issue. I was just a bit confused due to the documentation mentioning the syslinux script, i'm guessing the build-img script is basically the same as the syslinux script just with a different name?

fhs:
> thank you for your help with this issue. I was just a bit confused due to the documentation mentioning the syslinux script, i'm guessing the build-img script is basically the same as the syslinux script just with a different name?

Yeah, they do pretty much the same thing as far as I understand but implemented very differently. The `build-img.sh` script uses guestfish which mean it doesn't require root access and it'll also more portable since it doesn't depend on linux kernel's nbd module.

PR #1076, opened 2021-01-24
util/build-img.sh: use system syslinux binaries

fhs:
Guestfish uses the extlinux command installed on the system and the
version of syslinux install on the system may be different from the one
that comes with Harvey. If we mix syslinux versions, boot can fail. For
example, I was getting this when booting from an image created on an
ArchLinux system:
```
Undef symbol FAIL: x86_init_fpu
Failed to load libcom32.c32
Failed to load COM32 file vesamenu.c32
```

Help #1072

### DIFF
--- a/util/build-img.sh
+++ b/util/build-img.sh
@@ -2,18 +2,30 @@
 
 # Build a qcow2 image for loading in qemu
 
-SRC_IMG=$HARVEY/util/img/syslinux-bios
-SRC_SYSLINUX=$SRC_IMG/syslinux
-SRC_MBR=$SRC_IMG/mbr.bin
+# Syslinux should already be installed on the system since it's a
+# dependency of guestfish. We want to use the system syslinux binaries
+# which match the extlinux command below.
+SRC_SYSLINUX=/usr/lib/syslinux/bios
+SRC_MBR=/usr/lib/syslinux/bios/mbr.bin
+SRC_CFG=$HARVEY/util/img/syslinux-bios/syslinux/syslinux-usb.cfg
 SRC_KERNEL=$HARVEY/sys/src/9/amd64/harvey.32bit
 
-DEST=$HARVEY/harvey.qcow2
 
+FORMAT=qcow2
+if [ "$1" = "-f" ]; then
+	FORMAT=$2
+fi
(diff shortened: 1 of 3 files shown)


util/build-img.sh
```diff
@@ -2,18 +2,30 @@
 
 # Build a qcow2 image for loading in qemu
 
-SRC_IMG=$HARVEY/util/img/syslinux-bios
-SRC_SYSLINUX=$SRC_IMG/syslinux
-SRC_MBR=$SRC_IMG/mbr.bin
+# Syslinux should already be installed on the system since it's a
+# dependency of guestfish. We want to use the system syslinux binaries
+# which match the extlinux command below.
+SRC_SYSLINUX=/usr/lib/syslinux/bios
```
rminnich, line 8: I'd rather we bring the file into harvey. The reason is that this path is not at at all standard across distros. Would that be ok with you? These files change rarely if ever.

fhs: I'm ok with that. Any idea where the existing syslinux files were obtained from? I need the `extlinux` binary that matches `mbr` and other syslinux files already in harvey. Otherwise, I could just updated all syslinux files to the latest version.

gmacd: Might as well update all with the latest

fhs: I thought guestfish searches for `extlinux` in `$PATH`. I was wrong. It creates a VM image from the host system and runs everything from within qemu. So, modifying `$PATH` to include harvey's `extlinux` will not prevent guestfish from picking up the system `extlinux`. I don't see any easy workaround for this.

We can run `extlinux` externally after mounting the filesystem using `qemu-nbd` but that'd require root access and kind of defeats the purpose of using guestfish. Maybe we can just check for some common places linux distros keep the syslinux `*.c32` files?

Ultimately we want to provide downloadable boot images to users who want to try out harvey, so this script will be run by only few people.

rminnich (approved):
if you're still good with this let's go. I did not mean to slow it down.

rminnich (approved):

fhs:
At least the CI builds the image now. I'll go with this for now and improve it as things come up.

PR #1081, opened 2021-01-30
Add uroot service as a clone of terminal

gmacd:
The uroot service is a clone of terminal.  This can be the starting point to remove and tidy what we don't want.

If sys/src/9/amd64/allbuild.json is built (adding the uroot archive to the kernel ramdisk) it will unpack that to devramfs, bind it to root and bind /bbin to /bin, before other binaries.

To test:
- build everything
- build sys/src/9/amd64/allbuild.json
- run util/GO9PUROOT

### DIFF
--- a/rc/bin/urootrc
+++ b/rc/bin/urootrc
@@ -1,7 +1,7 @@
 #!/bin/rc
 # terminal startup
 
-if(test -e '#@/uroot') {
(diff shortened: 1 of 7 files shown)


rminnich (approved):
seems like a fine start.

Issue #1058, opened 2021-01-01
Add devdc() and devname() functions to all dev*.c


rminnich:
All dev*.c files should have these two functions
static char *
devname(void)
{
        return devtab.name;
}

static int
devdc(void)
{
        return devtab.dc;
}

That way, we avoid problems like the recent breakage with devregress.c where in the devtab we used 'R' but in the attach we use 'Z', which collided with devacpi

Good beginner issue. It might be a simple awk/sed script. For bonus points, use spatch.

michiboo:
Hi can I help on this issue?

gmacd:
Sorry for the delay @michiboo - any help is always appreciated!

PR #1082, opened 2021-01-31
Add devname() and devdc() functions to all dev*.c in sys/src/9/port

ecmma:
As per Issue #1058, I tried to add the two devname() and devdc() functions to all dev*.c (in sys/src/9/port and not in 9/arch). I don't know if the issue meant literally "devtab.name" as no variable devtab exists in such files; I added the functions using the dev-name, as in: 

static char *
devname(void)
{
        return tlsdevtab.name;
}

Sorry if I'm wrong, first PR and first approach to Harvey.

### DIFF
--- a/sys/src/9/port/dev9p.c
+++ b/sys/src/9/port/dev9p.c
@@ -648,3 +648,16 @@ Dev v9pdevtab = {
 	.remove = v9premove,
 	.wstat = v9pwstat,
 };
+
+static char *
+devname(void)
+{
+        return 9pdevtab.name;
+}
+
+static int
+devdc(void)
+{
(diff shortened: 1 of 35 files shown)


rminnich (commented):
I know it's a bit to ask, but if you could, further, explore how they should be used that would be good too. As it is you have added functions which are not used at all.

rminnich (commented):
don't forget your signed-off-by

What I'd suggest is, first, just doing one file, so we can go through how these functions are to be used.

ecmma:
Ok, sorry, I might have misunderstood the issue - I think I took it way too literally! I thought that how these functions were to be used was already well defined. I'll close the pull request.

PR #1085, opened 2021-02-06
Expose initrd file in devarch

gmacd:
If a file is passed to the kernel as a multiboot module (e.g. initrd), then that file is exposed by devarch.

Example:
- Add `--initrd uroot.cpio.lzma` to the qemu arguments
- Run harvey, and the file will be available to read under `#P/uroot.cpio.lzma`

Notes:
- Only the first initrd file is added to devarch
- If you pass a file in a subdirectory, the full path is used in the name of the file

### DIFF
--- a/sys/src/9/amd64/multiboot.c
+++ b/sys/src/9/amd64/multiboot.c
@@ -113,19 +113,11 @@ mbmoduleread(Chan *c, void *buf, i32 len, i64 off)
 	}
 
 	u64 modlen = initrd->modend - initrd->modstart;
-	void *data = (void*)(u64)initrd->modstart;
-
 	if (off < 0 || off > modlen) {
 		error("mbmoduleread: offset out of bounds");
 	}
-	print("mbmoduleread: initrd->data: %p initrd->len: %d buf: %p, len: %d, off: %lld\n",
-		data, modlen, buf, len, off);
-	if (off + len > modlen) {
-		len = modlen - off;
-		print("mbmoduleread: len adjust %d\n", len);
-	}
-	memmove(buf, data + off, len);
-	return len;
+	void *data = KADDR(initrd->modstart);
+	return readmem(off, buf, len, data, modlen);
 }
 
 int
@@ -194,13 +186,11 @@ multiboot(u32 magic, u32 pmbi, int vflag)
 			else {
 				u64 modstart = mod->modstart;
 				usize len = mod->modend - mod->modstart;
-				pamapinsert(modstart, len, PamMODULE);
+				pamapinsert(modstart, ROUNDUP(len, PGSZ), PamMODULE);
 
 				if (!initrd) {
 					addarchfile(p, 0444, mbmoduleread, nil);
 					initrd = mod;
-				} else {
-					print("Multiboot: can't add module %s to devarch\n", p);
 				}
 			}
 		}
--- a/sys/src/9/amd64/multiboot.c
+++ b/sys/src/9/amd64/multiboot.c
@@ -189,7 +189,13 @@ multiboot(u32 magic, u32 pmbi, int vflag)
 				pamapinsert(modstart, ROUNDUP(len, PGSZ), PamMODULE);
 
 				if (!initrd) {
-					addarchfile(p, 0444, mbmoduleread, nil);
+					// We just want the name - no directories
+					char *name = strrchr(p, '/');
+					if (!name || *++name == '\0') {
+						name = p;
+					}
+
+					addarchfile(name, 0444, mbmoduleread, nil);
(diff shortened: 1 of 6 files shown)


sys/src/9/amd64/multiboot.c
```diff
@@ -103,6 +105,29 @@ mbtypename(int type)
...
+mbmoduleread(Chan *c, void *buf, i32 len, i64 off)
+{
+	if (!initrd) {
+		error("mbmoduleread: module not found");
+	}
+
+	u64 modlen = initrd->modend - initrd->modstart;
+	void *data = (void*)(u64)initrd->modstart;
+
+	if (off < 0 || off > modlen) {
```
rminnich, line 118: see readmem, I think it does what you want.

sys/src/9/amd64/multiboot.c
```diff
@@ -164,17 +189,25 @@ multiboot(u32 magic, u32 pmbi, int vflag)
...
 				p = KADDR(mod->string);
 			if(vflag)
-				print("mod %#x %#x <%s>\n",
+				print("Multiboot: mod %#x %#x <%s>\n",
 				      mod->modstart, mod->modend, p);
 			else {
+				u64 modstart = mod->modstart;
 				usize len = mod->modend - mod->modstart;
-				pamapinsert(mod->modstart, len, PamMODULE);
+				pamapinsert(modstart, len, PamMODULE);
```
rminnich, line 197: did you check to make sure the pam code adds PTEs for this region.

gmacd: I'm not sure it is.  PamMODULE is not handled in meminit().  When I add it I see
```
panic: cpu0: mmukphysmap: remapping kernel direct address at va 0xffff800012a01044 (pml1 0xffff80000280a000 old 0x8000000012a01001 new 0x8000000012a01047)
cpu0 stack: k:ffff80000029e052 k:ffff800000248a22 k:ffff800000248b95 k:ffff800000246fdb k:ffff800000246487 ?:0
```

One thing I notice is that the area is not rounded to 4k.  Not yet sure if that's the particular problem here though.

sys/src/9/amd64/multiboot.c
gmacd: yup - thanks

sys/src/9/amd64/multiboot.c
```diff
@@ -103,6 +105,21 @@ mbtypename(int type)
...
 
+static i32
+mbmoduleread(Chan *c, void *buf, i32 len, i64 off)
+{
+	if (!initrd) {
+		error("mbmoduleread: module not found");
+	}
+
+	u64 modlen = initrd->modend - initrd->modstart;
+	if (off < 0 || off > modlen) {
```
rminnich, line 116: I think readmem should do the test of off < 0, and I think it already does the test for off > size.
So it makes sense (to me) to make sure readmem does the tests on lines 116-118 and then you can skip them here? 
for now I guess this will do.

sys/src/9/amd64/multiboot.c
```diff
@@ -164,17 +181,29 @@ multiboot(u32 magic, u32 pmbi, int vflag)
...
-				print("mod %#x %#x <%s>\n",
+				print("Multiboot: mod %#x %#x <%s>\n",
 				      mod->modstart, mod->modend, p);
 			else {
+				u64 modstart = mod->modstart;
 				usize len = mod->modend - mod->modstart;
-				pamapinsert(mod->modstart, len, PamMODULE);
+				pamapinsert(modstart, ROUNDUP(len, PGSZ), PamMODULE);
+
+				if (!initrd) {
```
rminnich, line 191: what if initrd is set? Isn't that some kind of error? What does this test mean? I think a comment would be helpful.

rminnich (commented):
very close :-)

sys/src/9/amd64/multiboot.c
```diff
@@ -103,6 +105,18 @@ mbtypename(int type)
 	return "(unknown)";
 }
 
+static i32
+mbmoduleread(Chan *c, void *buf, i32 len, i64 off)
+{
+	if (!initrd) {
```
rminnich, line 111: I suspect this test is not really absolutely necessary. This function won't get called without the addarchfile, and that only gets added if initrd exists. But no harm done I guess.

rminnich (approved):
Wow, we have initrd!

TODO: support more than one :-)

sys/src/9/amd64/multiboot.c
gmacd: True. I think it was a hangover from the original multi initrd support.

PR #1087, opened 2021-02-10
Remove allbuild.json, move u-root.json

gmacd:
Looks like I missed this commit yesterday.  With this change we no longer need allbuild.json - just build harvey normally, then build u-root if you want:

To build and run:
- bootstrap
- build
- build sys/src/u-root.json
- util/GO9PUROOT


sys/src/u-root.json
```diff
@@ -7,8 +7,8 @@
 			"GOARCH=amd64"
 		],
 		"Pre": [
-			"u-root -shellbang -initcmd= -o uroot.cpio plan9",
-			"lzma -f -k -3 uroot.cpio"
+			"u-root -shellbang -initcmd= -o /plan9_$ARCH/bin/uroot.cpio plan9",
```
fhs, line 10: Is there a better place to put this besides `bin`? It's not an executable binary. Maybe just `/plan9_$ARCH/uroot.cpio`?

fhs (approved):

sys/src/u-root.json
gmacd: fair point - changed

orangecms (approved):

PR #1089, opened 2021-02-10
Add prepare-commit-msg script

orangecms:
This will automatically sign off commits. See also:
https://stackoverflow.com/questions/15015894/git-add-signed-off-by-line-using-format-signoff-not-working/46536244#46536244


floren (approved):
You have to install it manually to make it happen, so it's not doing anything by default, and it's been tested, so here we go!

PR #1090, opened 2021-02-10
Add a script to fetch and compile apex, and bind in apex binaries in …

floren:
…the harvey profile

### DIFF
--- a/getapex.sh
+++ b/getapex.sh
@@ -1,0 +1,22 @@
+#!/bin/sh
+set -e
+
+# We assume you're running this from the Harvey dir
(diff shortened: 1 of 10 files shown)


getapex.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
```
rminnich, line 1: what shell is this?

rminnich (dismissed):

rminnich:
JOHN.

floren:
I always make a mess of signoffs. I'll try again.

PR #1091, opened 2021-02-10
Add a script to fetch apex automatically & build

floren:


bootstrap.sh
```diff
@@ -36,4 +36,8 @@ or to run in terminal mode:
   ./util/GO9PTERM
 
 To netboot, follow the instructions here: https://github.com/Harvey-OS/harvey/wiki/Booting-Harvey-on-real-hardware-I-(TFTP)
+
+You may want to run the following to get APEX installed:
+
+	OS=linux ARCH=amd64 CC=gcc sh getapex.sh
 EOF
```
rminnich, line 43: does this mean you are missing a newline at end of file?

usr/harvey/lib/profile
```diff
@@ -1,5 +1,6 @@
 bind -a $home/bin/rc /bin
 bind -a $home/bin/$cputype /bin
+bind -a /apex/amd64/bin /bin
```
rminnich, line 3: if this bind fails will the script break and stop here?

gmacd (dismissed):

usr/harvey/lib/profile
floren: No, I've tried it; an error is printed, but the script continues

bootstrap.sh
floren: Does what mean? That's a literal "EOF" string, it's a here document

floren:
ptal

rminnich (approved):

PR #1093, opened 2021-02-15
Pack dumped regs more tightly

gmacd:
This helps show more logging context when something goes wrong at boot and you have no serial


rminnich (approved):
wonderful. I wonder, since ds/es/fs/gs are vestigial could you pack them into one line ...
e.g
ds/es/fs/gs/cs/ss %#04x etc. etc
?
even use space not tab to force a fit?

gmacd:
They are on one line, tab separated.  E.g.
image

PR #1094, opened 2021-02-16
Copy uroot to tftp folder when built

gmacd:
To use, add `--- uroot.cpio.lzma` to the pxelinux file as described [here](https://github.com/Harvey-OS/harvey/wiki/Booting-Harvey-on-real-hardware-I-(TFTP)#preparing-the-host).

Note that at least on my machine, I get a page fault.  Fixing that would be the next step.  (Maybe it doesn't happen on other machines?)

### DIFF
--- a/clean.json
+++ b/clean.json
@@ -2,10 +2,11 @@
 	{
 		"Name": "clean",
 		"Pre": [
-			"rm -rf $ARCH/lib/*.a $ARCH/bin/*",
-			"rm -f cfg/pxe/tftpboot/harvey.32bit",
-			"rm -rf plan9_$ARCH/bin/*",
-			"rm -rf linux_$ARCH/bin/*"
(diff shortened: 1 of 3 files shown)


clean.json
```diff
@@ -4,6 +4,7 @@
 		"Pre": [
 			"rm -rf $ARCH/lib/*.a $ARCH/bin/*",
 			"rm -f cfg/pxe/tftpboot/harvey.32bit",
+			"rm -f cfg/pxe/tftpboot/uroot.cpio.lzma",
```
rminnich, line 7: quick question, should the cfg here be /cfg?
and same for the rest so this could work no matter where it's run from?

rminnich (approved):

clean.json
gmacd: Changed - could you re-review please?

floren (approved):

Issue #1096, opened 2021-02-17
harvey won't boot with initrd on real hardware


gmacd:
Repro steps:
- build harvey as usual
- build sys/src/u-root.json
- edit your pxe file ('default' in my case) and append `--- uroot.cpio.lzma` to the APPEND line)
- run centre, boot and see `#P/uroot.cpio.lzma` exists

If you have the same problem I had, it'll pagefault on boot.

The initrd memory area is recognised by multiboot, and identified as PamMODULE as expected.  It's placed after the kernel heap, but not immediately after - there's nearly 6Mi (5976064 bytes) of PamMEMORY in between the kernel heap and the module.

The problem seems to be in umeminit - when it's setting up the memory available for user space, it pagefaults when initialising the PamMEMORY area in between the kernel heap and the initrd module (in the plop function).

On real hardware, without the initrd module it boots fine.  Qemu boots fine with or without the initrd.

gmacd:
So I've got some ideas about the page fault with initrd on real hardware:

- In umeminit, setphysmembounds is called to set the user memory area to start from the start of the last PamMODULE section.
- umeminit then calls physinit to initialise every physical memory section of type PamMEMORY that's larger than 2Mi.
- The problem is that in my real hardware example, there's a PamMEMORY section larger than 2Mi between the kernel and the initrd module.  This is considered outside the physical memory bound (in setphysmembounds), but we still try to call physinit for it.

A quick fix would be to not call physinit for any memory section before the module sections (if there is one).  In this case you'd lose 4Mi of available ram (the actual size is just under 6Mi).

PR #1097, opened 2021-02-17
Avoid page fault due to partially initialised memory sections

gmacd:
If module memory sections exist (e.g. when using initrd), and there's a usable memory larger than 2Mi between the end of the kernel memory and the start of the module, harvey would try to use it as user memory, despite it being found to be earlier than sys->pmstart (as set in setphysmembounds).  This would cause a page fault when physinit was later called for that memory section.

This commit will avoid using any PamMEMORY sections found earlier than any PamMODULE sections.  This has the downside that we lose useful memory, but probably not very much.  In my case the memory section was just under 6Mi.

Maybe this can be recovered with a more sophisticated change (probably involving physalloc.c)?

This change should have no effect if modules (initrd) are not used.

### DIFF
--- a/sys/src/9/amd64/memory.c
+++ b/sys/src/9/amd64/memory.c
@@ -96,7 +96,20 @@ umeminit(void)
 	extern void physallocdump(void);
 	setphysmembounds();
 
+	// If there are any modules, start user memory after the last module.
+	// This avoids any large PamMEMORY sections between the kernel and the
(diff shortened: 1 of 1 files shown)


gmacd:
Fixes #1096

rminnich (approved):
not sure I understand it but if it works ... it works.

floren:
I've tested the change on an apu2 and it boots ok, so let's give it a shot.

PR #1101, opened 2021-04-11
Replace size_t with usize; YYSIZE_T is now usize

rminnich:

### DIFF
--- a/aarch64/include/u.h
+++ b/aarch64/include/u.h
@@ -22,7 +22,7 @@ typedef unsigned long long u64;
 typedef u64 uintptr;
 typedef u64 uintptr_t;
 typedef u64 usize;
-typedef u64 size_t;
+typedef i64 isize;
 typedef u32 Rune;
 
 typedef union FPdbleword FPdbleword;
--- a/riscv/include/u.h
+++ b/riscv/include/u.h
@@ -24,8 +24,6 @@ typedef unsigned int	uint;
 typedef i64 intptr_t;
 typedef u64 uintptr;
 typedef u64 uintptr_t;
-typedef u32	usize;
-typedef i64 size_t;
 typedef	u32		Rune;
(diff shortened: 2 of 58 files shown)


aarch64/include/u.h
```diff
@@ -22,7 +22,7 @@ typedef unsigned long long u64;
 typedef u64 uintptr;
 typedef u64 uintptr_t;
 typedef u64 usize;
-typedef u64 size_t;
+typedef u64 usize;
```
gmacd, line 25: isn't this a duplicate?

riscv/include/u.h
```diff
@@ -25,7 +25,7 @@ typedef i64 intptr_t;
 typedef u64 uintptr;
 typedef u64 uintptr_t;
 typedef u32	usize;
-typedef i64 size_t;
+typedef i64 usize;
```
gmacd, line 28: also dupe and conflicting type

aarch64/include/u.h
rminnich: fixed, thanks!

riscv/include/u.h
rminnich: I'm not going to touch riscv just yet. I doubt it builds.

rminnich: actually, I agree with you and changed it.

gmacd: tbh i think it might be worth just deleting all the riscv code and maybe the aarch64 code too.  was the progress made for these ports worth the weight and time required to 'maintain' them?

gmacd (approved):

riscv/include/u.h
dancrossnyc: I think we should revive aarch64 and riscv; regardless, what we do there is out of scope for this change.

I agree that usize should be u64, but I would consider keeping around a typedef for size_t just got compatibility.

PR #1103, opened 2021-04-13
replace intptr_t with isize

rminnich:


rminnich:
I've tested this but you can too.

floren (approved):
Well, it compiles and boots, so I'm not complaining.

PR #1104, opened 2021-04-13
remove uint

rminnich:
It's got an implied size (32 bits) and that was always confusing.

Use u32.

Per Charles, for Plan 9 source and kenc, it was best to make
everything *except* long long be 32 bits, including int.
This is the same as gcc as it happens.


rminnich:
As for int ... that one is probably ok to leave alone. But uint always confused at least me.

gmacd (approved):

PR #1105, opened 2021-04-14
rename uintmem to physaddr

rminnich:
It was named uintmem because uint is 32-bits and we needed a
physical address type that was kind of uint but really for physical
addresses and hence 64 bits, while, at the same time, not being
tied to a bit length (so u64 would not do).

This all goes back to the days of the 32-bit transition for K10 IIRC.

This is all very confusing so give it a name related to its function.

### DIFF
--- a/sys/src/9/amd64/archamd64.c
+++ b/sys/src/9/amd64/archamd64.c
@@ -475,7 +475,7 @@ archfmtinstall(void)
 	 * Architecture-specific formatting. Not as neat as they
 	 * could be (e.g. there's no defined type for a 'register':
 	 *	L - Mpl, mach priority level
-	 *	P - uintmem, physical address
+	 *	P - physaddr, physical address
 	 *	R - register
 	 * With a little effort these routines could be written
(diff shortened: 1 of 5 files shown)


gmacd (approved):

dancrossnyc:
May I recommend calling this `u64` instead? We found when we needed an integral type to hold physical addresses that made the most sense (aside from newtype'ing the thing in Rust). Our rationale was that a physical address is just a 64-bit integer, as opposed to a virtual address, that could be represented in a usize.

rminnich:
Dan, reason I went with physaddr was that jmk, or charles, not sure who, long ago, felt it was important to distinguish that small set of uses/definitions. 

Happy to revisit this at some point but I'd rather let it stay distinguished for now ... I have a feeling they had a reason for it. But maybe not, who knows.

rminnich:
Dan, I have asked Charles about this, and depending on the answer, we can see if your suggestion makes more sense. 

I feel happy having someone with your experience look at this stuff and raise the questions, many thanks.

PR #1111, opened 2021-04-23
Define rules for importing code.

rminnich:


gmacd (approved):
In fairness the example of a breaking PR was a WIP, but otherwise the doc looks good.

PR #1112, opened 2021-04-23
Add 9front nusb

gmacd:
Under sys/src/cmd/frontusb, following in the footsteps of legacyusb.
Added LICENSE file in same directory, with contents copied from /lib/legal/mit in 9front as indicated in the FQA: http://fqa.9front.org/fqa0.html
First commit, doesn't build.

### DIFF
--- a/sys/src/cmd/frontusb/LICENSE
+++ b/sys/src/cmd/frontusb/LICENSE
@@ -1,0 +1,20 @@
+Copyright © 2021 Plan 9 Foundation
+Copyright © 20XX 9front authors
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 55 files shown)


sys/src/cmd/frontusb/LICENSE
```diff
@@ -0,0 +1,20 @@
+Copyright © 2021 Plan 9 Foundation
+Copyright © 20XX 9front authors
```
floren, line 2: Time for some annoying license nit-picking: Since this comes from their boilerplate license file, it just says "20XX". I'm not sure what's exactly correct, but there should probably be a year in here. Maybe 2021, since we imported it in 2021? Otherwise you could look at the git history of the directory and see when they last modified it?

floren (approved):

PR #1123, opened 2021-05-02
Dev create signature changed to return Chan*

gmacd:
Device create functions changed to return Chan* created, similar to open.
Required for integration of devshr.

Looks to have been added because shrcreate has a path where a new Chan is created, in which case it wants to sent the new chan rather than the one passed in.

As mentioned in slack, the 9front commit this change is based on is: https://code.9front.org/hg/plan9front/rev/2e0310025a75
```
changeset:   771:2e0310025a75
user:        cinap_lenrek@centraldogma
date:        Wed Aug 17 23:27:31 2011 +0200
summary:     change definition of Chan.create to return a chan like open
```

### DIFF
--- a/sys/src/9/ip/devip.c
+++ b/sys/src/9/ip/devip.c
@@ -513,10 +513,11 @@ ipopen(Chan *c, int omode)
 	return c;
 }
 
-static void
+static Chan*
 ipcreate(Chan *c, char *n, int i, int m)
 {
 	error(Eperm);
+	return nil;
 }
 
 static void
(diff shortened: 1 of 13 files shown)


rminnich (changes requested):
we need to know *why* that change was made. You've just created an issue where a function that returned void is not returning something, and all the call sites will ignore the return value. 

Understanding must precede changing.

gmacd:
I've updated the description to reference the 9front commit this change is based on, and the reason behind it.

rminnich (approved):

PR #1134, opened 2021-07-23
Implement some really, really basic Github Actions stuff

floren:
that boots from the 9legacy ISO and runs a command. For use
as a jumping-off point.

### DIFF
--- a/.github/workflows/harvey.yml
+++ b/.github/workflows/harvey.yml
@@ -10,8 +10,6 @@ jobs:
         working-directory: build
     steps:
     - uses: actions/checkout@v2
-      with:
-        lfs: 'true'
     - name: install deps
       run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/harvey.yml
```diff
@@ -0,0 +1,25 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: build
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        lfs: 'true'
```
gmacd, line 14: guess we don't need this now

gmacd (approved):

floren:
PTAL

dancrossnyc (approved):

PR #1135, opened 2021-07-24
Make the workflow install & run ufs, then mount it and do an ls.

floren:
Now we're booted into an environment with all the tools we need to build, and the source we intend to build from in /n/harvey. Baby steps.

Signed-off-by: John Floren <john.floren@gravwell.io>


gmacd (approved):

sevki (changes requested):
probably should run on pullrequests as well, not just pushes :smile:

floren:
PTAL

sevki (approved):

PR #1136, opened 2021-07-24
merge harvey-go into harvey

rminnich:
we are going to make Go a first class citizen, might as well bring in harvey-go.


dancrossnyc (approved):

gmacd:
I'd argue we don't necessarily want all the go programs in tree.  Some yes, but only those that are generally useful.  I'd also be wary of the ipfs cmd - it's really cool, but IIRC it takes ages to build.

sys/postpipe.go
```diff
@@ -0,0 +1,34 @@
+// Copyright 2020 The Ninep Authors.  All rights reserved.
```
floren, line 1: I don't think we want this in /sys/postpipe.go

README.md
```diff
@@ -0,0 +1 @@
+# Go Packages
```
floren, line 1: we should get rid of this README file, it'll only conflict with the real one

sys/src/cmd/preen/.gitignore
```diff
@@ -0,0 +1 @@
+/preen
```
floren, line 1: this isn't right

rminnich: why?
it's saying the preen binary in this directory should be ignored.

sys/postpipe.go
rminnich: moved do sys/src/cmd/postpipe.go

README.md
rminnich: removed

sys/src/cmd/preen/.gitignore
rminnich: If the pattern contains a slash, which is not in the trailing position (it's not the last character), then it is matched only with pathnames relative to the .gitignore file location.

floren (approved):

PR #1142, opened 2021-07-26
Run go command in command.bash, check version

rminnich:


build/command.bash
```diff
@@ -31,6 +31,12 @@ send "9fs tcp!10.0.2.2!5640 /n/harvey\n"
 expect -exact "term% "
 send "ls -l /n/harvey/\n"
 
+# check our Go version
+send "bind -a /go/bin /bin\n"
```
floren, line 35: You should really put a `expect -exact "term% "` before you send the bind command to keep things clean.

gmacd (approved):

PR #1143, opened 2021-07-31
License pivot sprint branch

officialplan9foundationofficial:


rminnich:
I don't understand the point, and the comment is not useful, and it's not signed off; closing it.

PR #1144, opened 2021-08-02
Import git9

dancrossnyc:
Thanks, Ori!


oridb:
Just want to point out that this change will also need a few changes that don't seem to have made it in to the repo.
They're mentioned in the readme:

	- rc-line-split: implement delim{...} syntax for rc
	- walk: port walk command to 9legacy.
	- aux/getflags: 9front updated it with support for named args

There are a few others that probably got in as git9 evolved; eg, off the top of my head, we added
the `diff -u` flag to output unified diffs from the base system, and I think we use that in a few places,
and I think we parse dates using the new formatted date code from tmdate(2):

	http://man.9front.org/2/tmdate

There's a crude regression that's starting to grow here, which is probably worth running:

	http://shithub.us/ori/regress/git/cmd/git/git.rc/f.html

Feel free to add the tests to harvey and subscribe me to the output once you get them
running (and, if there are tests that belong there, I'd be happy to add them)

dancrossnyc:
Thank you, Ori; this is very helpful.

oridb:
Oh -- as a final note, while `git/compat` is present for go, it can probably use a bit of love to stub out a few more commands, mostly around tags.

floren:
Dan, can you set the BUGGERED variable in /sys/src/cmd/mkfile to `BUGGERED=unix|git`? This should stop it from being included in `mk all` and breaking builds. Then the code is in the tree and we can iterate through the necessary patches (we already have walk in, apparently, but not the other two).

floren:
Another thing we need: bring in Bfdopen. You should be able to just copy it from /sys/src/cmd/walk.c, or we can promote it to libbio as 9front appears to have done.

dancrossnyc:
> Another thing we need: bring in Bfdopen. You should be able to just copy it from /sys/src/cmd/walk.c, or we can promote it to libbio as 9front appears to have done.

We should promote it; I see no reason not to.

dancrossnyc:
> > Another thing we need: bring in Bfdopen. You should be able to just copy it from /sys/src/cmd/walk.c, or we can promote it to libbio as 9front appears to have done.
> 
> We should promote it; I see no reason not to.

(However, that should be a separate change.)

gmacd (approved):

PR #1153, opened 2021-08-11
git: change time-related log code to use our libc rather than 9front (fix git build)

gmacd:
Changes formatting of long dates in git log.  Shouldn't make a difference.  Sufficient to get git to build and run.

### DIFF
--- a/sys/src/cmd/git/log.c
+++ b/sys/src/cmd/git/log.c
@@ -135,7 +135,7 @@ nextline(char *p, char *e)
 static void
 show(Object *o)
 {
-	Tm tm;
+	Tm *tm;
 	char *p, *q, *e;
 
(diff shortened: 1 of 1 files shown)


sys/src/cmd/git/log.c
```diff
@@ -325,7 +325,7 @@ main(int argc, char **argv)
 		sysfatal("chdir: %r");
 
 	gitinit();
-	tmfmtinstall();
+	//tmfmtinstall();
```
dancrossnyc, line 328: If we're making this change, then delete this file: git remembers so that we don't have to.

But it may make sense to reach out to Ori and collaborate on a shim library for 9front compat?

rminnich (approved):

PR #1156, opened 2021-08-14
Add support for bringing in Go

rminnich:
if GOLANG_HARVEY is set, the script does a bin mount to sys/go

from there, go will be included.

Tested and "just works".

The amount of memory for qemu had to be bumped to 4096m.

Depending on what is in $GOLANG_HARVEY, the CD may or may not be
big enough.

### DIFF
--- a/build/build.bash
+++ b/build/build.bash
@@ -1,6 +1,23 @@
 #!/bin/bash
 set -e
 
+rm -f /tmp/image
+
+if [ $HARVEY_GOLANG"" == "" ]; then
+               echo Not including Go in the image
+       else
+	      mkdir -p ../sys/go
+              echo sudo mount --bind $HARVEY_GOLANG ../sys/go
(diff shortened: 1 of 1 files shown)


build/build.bash
```diff
@@ -1,6 +1,23 @@
 #!/bin/bash
 set -e
 
+rm -f /tmp/image
+
+if [ $HARVEY_GOLANG"" == "" ]; then
+               echo Not including Go in the image
```
gmacd, line 7: Strange tabbing. Once we have a build of go available we should probably make it non optional.

gmacd (approved):

build/build.bash
rminnich: we'll clean this up later I guess. I'd be way happier if this script were a Go program, I don't like building in bash dependencies.

PR #1158, opened 2021-08-16
Fix git9 build in the distribution.

paul-lalonde:
Git was being compiled but its script library was not included in the distribution.

Add /sys/lib/git to mkdirs
Add /sys/lib/git to dist/replica/plan9.proto


rminnich (dismissed):

paul-lalonde:
Abandoning so I can learn this flow better.  Sorry for the churn.

PR #1164, opened 2021-09-03
Split CDROM and USB building into separate scripts. Also clean up bui…

floren:
…ld stuff a bit.

Even the Go bootstrap is 312MB once unpacked. An iso without Go is 454
MB already.  That's 766MB total. If we're going to keep ISO building
around, we're probably going to have to do it without shipping Go,
unless we want to tell everyone to write to DVD.

Signed-off-by: John Floren <john.floren@gravwell.io>


rminnich (approved):
looks fine but I'm not sure .iso size makes this change essential. Plus, every change we do we'll now need to check two files. OTOH breaking them out makes it easier to do the one thing you want instead of a thing you want and a thing you don't want ... and I only want usb anyway :-)

PR #1170, opened 2021-09-18
Tweaks to startup - better display defaults, start edwood instead of acme

gmacd:
Try to improve harvey startup by providing better default display settings, running edwood on startup.
Also change build expect strings to regexps so it can work with different defaults in the future (when we build with harvey 0.0.1 rather than 9legacy).


usr/glenda/bin/rc/riostart
```diff
@@ -13,5 +13,5 @@ if not
 	dump=acme.dump
 
 a=`{echo $wid-35 | hoc }
-window 60,90,$a,$ht acme -l lib/$dump
+window 60,90,$a,$ht edwood -l lib/$dump
```
floren, line 16: It's currently possible to build Harvey without Go; in fact, I haven't yet figured out how to build an ISO *with* Go (mk9660 chokes in a weird way I've been unable to reproduce artificially).

I suppose if edwood doesn't exist, we just won't get any editor window at all, which is fine.

gmacd: Ah well, I’ll remove it for now and we can add it again later

rminnich: it's fine, acme just won't run, don't sweat it. no harm done.

This is fine.

rminnich (approved):

PR #1171, opened 2021-09-20
Unixtoolchain

rminnich:
This builds a an untested toolchain that runs on unix that can be used to see if a kernel will even build.

Doesn't quite work yet; there was more tweaking we did on NxM to make all the pieces go.

Leave it here, as a branch, separate repo? Your call.


dancrossnyc:
I think separate repo makes the most sense here. NxM is a bit different as we can diff against it using git tools.

rminnich:
This is here for reference but I've got a better way to do this, see #1182

PR #1172, opened 2021-09-21
WIP: change rdmsr to be safe

rminnich:
Add labels in rdmsr to indicate which path to take on a failure.

Add GPF handler.

In GPF handler, switch on ureg->pc value, and if an address is handled,
fix uref->pc.

This is basically what is done in linux, minus all the fancy ELF and
linker set machinations. It will work just fine.

### DIFF
--- a/sys/src/9/pc/l.s
+++ b/sys/src/9/pc/l.s
@@ -671,9 +671,19 @@ TEXT lcycles(SB),1,$0
 	RDTSC
 	RET
 
+// WIP: if the rdmsr fails, then return all -1
+// TODO: change type signature of rdmsr
 TEXT rdmsr(SB), $0				/* model-specific register */
 	MOVL	index+0(FP), CX
+TEXT rdmsr_doit(SB), $0
 	RDMSR
+TEXT rdmsr_ok(SB), $0
+	JMP 1f
+TEXT rdmsr_bad(SB), $0
+	MOVL    $0xffffffff, AX
+	MOVL	AX, DX
+
+1:
 	MOVL	vlong+4(FP), CX			/* &vlong */
 	MOVL	AX, 0(CX)			/* lo */
 	MOVL	DX, 4(CX)			/* hi */
--- a/sys/src/9/pc/trap.c
+++ b/sys/src/9/pc/trap.c
@@ -221,6 +221,7 @@ trapinit(void)
 	 */
 	trapenable(VectorBPT, debugbpt, 0, "debugpt");
 	trapenable(VectorPF, fault386, 0, "fault386");
+	trapenable(VectorGPF, faultgpf, 0, "faultgpf");
 	trapenable(Vector2F, doublefault, 0, "doublefault");
 	trapenable(Vector15, unexpected, 0, "unexpected");
 	nmienable();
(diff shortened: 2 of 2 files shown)


sys/src/9/pc/l.s
```diff
@@ -671,9 +671,19 @@ TEXT lcycles(SB),1,$0
...
 	RET
 
+// WIP: if the rdmsr fails, then return all -1
+// TODO: change type signature of rdmsr
 TEXT rdmsr(SB), $0				/* model-specific register */
 	MOVL	index+0(FP), CX
+TEXT rdmsr_doit(SB), $0
 	RDMSR
+TEXT rdmsr_ok(SB), $0
+	JMP 1f
```
paul-lalonde, line 681: I don't know the assembler particularly - is this JMP correct?  Your label is 1, but the JMP to 1f.

rminnich: we're sending you back to school. This is a Knuth invention, in Unix assemblers for 50 years. :-)

https://stackoverflow.com/questions/27353096/1b-and-1f-in-gnu-assembly

it's incredibly handy.

rminnich: I'll go ahead and push this if tests show it does no harm

sys/src/9/pc/trap.c
```diff
@@ -608,6 +609,21 @@ unexpected(Ureg* ureg, void*)
 
 extern void checkpages(void);
 extern void checkfault(ulong, ulong);
+
+static void
+faultgpf(Ureg* ureg, void*)
+{
+	switch (ureg->pc) {
```
dancrossnyc, line 616: Start small and use an `if` here?

```
        if(ureg->pc != rdmsr_doit)
                panic("GPF with no handler at addr 0x%.8lux", ureg->pc);
        ureg->pc = rdmsr_bad;
```

sys/src/9/pc/l.s
```diff
@@ -671,9 +671,19 @@ TEXT lcycles(SB),1,$0
 	RDTSC
 	RET
 
+// WIP: if the rdmsr fails, then return all -1
+// TODO: change type signature of rdmsr
 TEXT rdmsr(SB), $0				/* model-specific register */
 	MOVL	index+0(FP), CX
+TEXT rdmsr_doit(SB), $0
 	RDMSR
+TEXT rdmsr_ok(SB), $0
```
dancrossnyc, line 680: I don't think you need this label. But I'd probably do the inverse; assume success, and have the failure case after tha mainline code, and then have it jump back to  the exit:

```
TEXT rdmsr(SB), $0
        MOVL    index+0(FP), CX
TEXT mayberdmsr(SB), $0
        RDMSR
1:
	MOVL	vlong+4(FP), CX			/* &vlong */
	MOVL	AX, 0(CX)			/* lo */
	MOVL	DX, 4(CX)			/* hi */
	RET
TEXT rdmsrfail(SB), $0
	MOVL    $0xffffffff, AX
	MOVL	AX, DX
        JMP     1b
```

rminnich:
This works, I like the switch as opposed to the if, the assembly is basically the same and the switch gives a pattern for future handlers; but I like dan's assembly code and will probably change to match his suggestion. But the next step is to change the name and type signature of rdmsr, so I'll make those tweaks in the next pass. Thanks everyone for the reviews!!

PR #1176, opened 2021-09-23
Fixed rdmsr.

dancrossnyc:
Our assembler doesn't understand number labels
and the 'f' and 'b' syntax.  Case labels in a
switch must refer to compile-time constants.


rminnich (approved):
Thanks.

PR #1179, opened 2021-09-24
k10: integrate parts of Geoff's kernel

dancrossnyc:
There's enough of Geoff's kernel in here now not just
panic on startup.  With the 9front 82563 driver, it
recognizes my ethernet interfaces, though something
still fails.


sys/src/9k/k10/mem.h
```diff
@@ -60,6 +60,8 @@
 #define USTKSIZE	(16*1024*1024)		/* size of user stack */
 #define TSTKTOP		(USTKTOP-USTKSIZE)	/* end of new stack in sysexec */
 
+#define	KSEG0SIZE	(256ull*MiB)
```
rminnich, line 63: dang. KSEG0 again.

rminnich (approved):

PR #1182, opened 2021-10-02
Allow building 6l on unix

rminnich:
This is the first in a series of scripts that will let us build the toolchain on linux.

Adapted from NxM, but better: it's just two files for 6l, and it's buildable right in the tree.

### DIFF
--- a/sys/src/cmd/6l/BUILD
+++ b/sys/src/cmd/6l/BUILD
@@ -1,0 +1,16 @@
+#!/bin/sh
+mkdir -p ../../../../linux_amd64/bin
+9c -Wno-char-subscripts -g -I .  asm.c list.c obj.c optab.c pass.c span.c unixcompat.c ../6c/enam.c ../ld/elf.c
+
+9l -o 6l \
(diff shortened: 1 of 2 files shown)


rminnich:
doing this in parallel with watching daniel's podcast :-) It's a really good podcast!

gmacd (approved):

gmacd:
Any particular reason the build script is called BUILD?

sys/src/cmd/6l/BUILD
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+mkdir -p ../../../../linux_amd64/bin
```
dancrossnyc, line 2: I'd greatly prefer that this just creates a binary called `6l.linux_elf64` or something and then we can copy it wherever needed.

dancrossnyc:
> Any particular reason the build script is called BUILD?

Presumably because it builds the program. :-) But calling it `mklinux` or something would be more in keeping with the existing style.

PR #1194, opened 2021-11-22
DRAFT: Add support for mips64 building

rminnich:
4c doesn't build yet, the others do. This was done in about 4m so
likely not quite there.


dancrossnyc (commented):
I'm happy to see other architectures get built, but what I'm less happy about is all these BUILD scripts scattered all over the system, particularly since they seem to have embedded knowledge of, e.g., where they exist in the FS hierarchy.

Given that this is all for cross-compilation, I wonder if perhaps we can put them in their own little subdirectory?

rminnich:
let's dump it

PR #1199, opened 2021-11-22
Add legacy ?.out.h files to libmach

gmacd:
After removing the old 32 bit toolchains (sparc, power, mips), we need to move the ?.out.h files into libmach and modify the mkfile to make things build again.


rminnich (approved):
we have got to get our CI working :)

PR #1202, opened 2021-11-29
Simplify 4c mkfile - move 4c/lex.c changes to cc/lex.c

gmacd:
4c/lex.c is near identical to cc/lex.c, but misses a couple of recent fixes, and has a little 4c-specific code (mips64).

cc/lex.c already has some vc-specific code in it (mips32), so we can get rid of 4c/lex/c by moving that code to cc/lex.c

The reason to do this is to make 4c easier to build on linux (due to the hack to use 4c/lex.c instead of cc/lex.c when building).


sys/src/cmd/4c/mkfile
```diff
@@ -29,9 +29,7 @@ enam.c:	4.out.h
 	rc mkenam
 
 $LIB:
-	rfork
```
rminnich, line 32: this worries me. 

which one is 4 again?

sys/src/cmd/cc/lex.c
```diff
@@ -60,12 +60,21 @@ main(int argc, char *argv[])
 		break;
 
 	case 'l':			/* for little-endian mips */
-		if(thechar != 'v'){
-			print("can only use -l with vc");
+		switch(thechar) {
```
rminnich, line 63: ah, why is this? I am now lost. The commit message is not helping at all.

sys/src/cmd/4c/mkfile
gmacd: 4 is mips64.  4c has a custom version of lex.c  this part of the makefile differs from other ?c mkfiles because of this.  this change makes the 4c mkfile consistent with other ?c mkfiles.

the only change made in 4c/lex.c that hasn't been made in cc/lex.c is related to some already-mips-specific code.  since there's already mips-specific code in cc/lex.c, I've made the change in this PR.

we can make this change because in this PR we take the

sys/src/cmd/cc/lex.c
gmacd: if you diff 4c/lex.c and cc/lex.c you'll see this is the main difference, so i moved it to cc/lex.c.

of course this could all go away if we can agree mips is dead....

rminnich:
you'll need to rebase this and then confirm that you tested it with build-usb.bash, earlier removals broke native building and I had to revert them

PR #1205, opened 2021-12-31
Add rc to boot options, and get rid of 9load usb boot dependency.

floren:
You can now pick "rc" when asked "root is from". This gives you an rc
shell to set stuff up.

This change also invokes disk/fdisk and disk/prep when booting from
a USB disk. Previously, we had received partition offsets from 9load
and passed them in via the sdB0part environment variable. Since nload
doesn't do that, and we'd like to move to nload, we needed a different
way. Note that we had to add a few additional programs to /boot (namely
rc, fdisk, and prep) to make this work, but the kernel is still small.


rminnich (approved):
NIce to see this coming in; it was only reduced to practice 15+ years ago, for linuxbios, but things take time in Plan 9 :-)

rminnich (approved):

PR #1206, opened 2021-12-31
nboot: add nboot and its efi booters to the build process

rminnich:
Next is getting it on the 9fat.


rminnich:
john, format only puts stuff in at top level, right? suck. Guess I'll have to mount the partition. Suck.

floren (approved):

floren:
As for `format`, try creating the efi/boot directory and populating it with the files, then specify *the efi directory* as one of the arguments to format? Might work. Might not. Worth trying, though.

PR #1210, opened 2022-01-26
rdmsr: Don't panic on user GPF.

dancrossnyc:
I believe this fixes what David observed.

### DIFF
--- a/sys/src/9/pc/trap.c
+++ b/sys/src/9/pc/trap.c
@@ -327,7 +327,7 @@ trap(Ureg* ureg)
 	}
 
 	m->perf.intrts = perfticks();
-	user = (ureg->cs & 0xFFFF) == UESEL;
+	user = userureg(ureg);
 	if(user){
 		up->dbgreg = ureg;
 		cycles(&up->kentry);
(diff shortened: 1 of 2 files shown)


gmacd (approved):
This looks fine to me, any reason not to approve and merge?

dancrossnyc:
Some testing would be nice.... :-D

sys/src/9/pc/trap.c
```diff
@@ -616,6 +616,13 @@ extern void rdmsrfail(void);
 static void
 faultgpf(Ureg* ureg, void*)
 {
+	int user = (ureg->cs & 0xFFFF) == UESEL;
```
rminnich, line 619: There's a macro in fns.h
fns.h:#define   userureg(ur) (((ur)->cs & 0xFFFF) == UESEL)
might want to use that.

dancrossnyc: Sure. Fixed up a few other locations of that sequence while I was at it; I left one (which is in a large expression also looking for data segments).

rminnich (approved):

PR #1217, opened 2022-03-30
Updating commands doc to use https instead of git

Doom4535:
This is do to the recent github change removing support for the unauthenticated git protocol on port 9418. For more info, see: https://github.blog/2021-09-01-improving-git-protocol-security-github/

If people depended on `git://` because of poor network connections, this may impact it (may want to switch to using `ssh://` in those cases).


Doom4535:
I noticed this while trying to setup a build environment (I have not succeeded at building this yet).

orangecms (approved):

orangecms:
Yea, thank you! There is also a Plan 9 git implementation that we need to check if it works with HTTPS, I guess.

rminnich:
welcome to harvey :-)

PR #1218, opened 2022-04-18
Add the first (more to come) paper from the DOE Fast-OS work

rminnich:
Signed-off-by: Ronald Minnich <rminnich@gmail.com>


orangecms (approved):

orangecms:
Do you have the individual files? We could split the PDF into the various papers within. Then we can name them accordingly.

rminnich:
You mean the source to the paper? It's long gone.

orangecms:
> You mean the source to the paper? It's long gone.

No, I scrolled through and saw that it's actually 
 6 papers in one file, not just what the file name suggests; in addition:

- Easy system call tracing for Plan 9
- Recent Plan 9 Work at Bell Labs
- A Cache to Bash for 9P
- HTTP-Like Streams for Plan 9
- Effective Resonant Frequency Tracking With Inferno

rminnich:
oh wow! sorry! academia ran that one right by me. Let's split them out, apologies!

